PR #2, opened 2017-10-02
Merge setup_vars_*.sh into env.sh

maruel:
This reduces the different flows users have to follow based on their OS.

Remove the +x bit on the script, since it is not meant to be executed but
sourced. Remove the shebang for the same reason.


maruel:
I guess "setupvars.sh" could be fine too but I'm more used to env.sh, as you prefer.

deadprogram:
I like it, and thank you! Now merging.

PR #3, opened 2017-10-04
Consolidated gocv command line tool into one executable

ulisesflynn:
This is a first cut into consolidating all the gocv commands into one executable, but open to any improvements/suggestions.


deadprogram:
I'm not sure this is what @mattetti had in mind when suggested turning the examples into commands. I was thinking each command did a single thing, not creating a full CLI for OpenCV. But I'm open to posibilites. Anyone else want to chime in here?

gedw99:
I like it...
In fact its a great Test Harness. Great to validating stuff and new users also getting the hang of the API:

mattetti:
That's exactly what I had in mind, it's a great entry point, they should get build when the tests are running to verify they still compile. When browsing projects like that, I am always looking for an example I can easily run and then tweak for my own needs. I thinks those commands can be exactly that.

mattetti:
Clarification: I was responding to Ron, I do think we are better off with multiple small commands that can be used as examples and starting points. My understanding is that this is a library first, not a final standalone product so I would probably suggest to create a global CLI in a separate project relying on this library.

deadprogram:
The question is do you want them built as a single command executable, or as a group of small command executables, is I think the key question here. I tend towards the latter.

mattetti:
yes me too, small command executables seem better to me. That shouldn't
prevent someone else to create a single global executable program outside
of this repo tho, tight?

ulisesflynn:
I think that with the intention of this being a library, the small command executables is fine to show usage of the library, since it's not a standalone application. If everyone is good with it, I'll close this request.

PR #12, opened 2017-10-18
Add Resize.

kellegous:
This one adds support for `cv::resize`. I left off a couple of values from the `InterpolationFlags` enum for now since they are only used in other geometric transformations.


deadprogram:
Tremendous, thank you! Merging.

PR #13, opened 2017-10-19
Suggest install of dep github.com/saljam/mjpeg

drnic:


drnic:
Awww, the spacing isn't right. Closing and will do it properly from local machine.

PR #14, opened 2017-10-19
suggest mjpeg-stream dep github.com/saljam/mjpeg

drnic:


deadprogram:
Thank you @drnic for the correction. By the way, I switched over the PR to the `dev` branch before merging.

PR #19, opened 2017-10-28
Add image compression flags for gocv.IMWrite

lutskanu:
see
https://docs.opencv.org/3.3.0/d4/da8/group__imgcodecs.html#gabbc7ef1aa2edfaa87772f1202d67e0ce
https://docs.opencv.org/3.3.0/d4/da8/group__imgcodecs.html#ga292d81be8d76901bff7988d18d2b42ac

### DIFF
--- a/cmd/saveimage/main.go
+++ b/cmd/saveimage/main.go
@@ -49,5 +49,5 @@ func main() {
 		return
 	}
 
-	gocv.IMWrite(saveFile, img, []int{})
+	gocv.IMWrite(saveFile, img)
 }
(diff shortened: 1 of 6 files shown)


deadprogram:
Hi @lutskanu thank you for working on this. My only thought, is perhaps we should have both `IMWrite()` and also `IMWriteWithParams()` or else possibly a single function but use a variadic param.

What do you think?

lutskanu:
Hi @deadprogram I think you are right. Will fix it today

deadprogram:
Thanks, @lutskanu for the revision to your PR. Now merging.

PR #21, opened 2017-11-02
return temp pointer fix

longlene:
ostringstream.str() returns a temporary string object that's destroyed at the end of the function.
return the macro is just enough.

### DIFF
--- a/version.cpp
+++ b/version.cpp
@@ -1,12 +1,5 @@
 #include "version.h"
-#include <sstream>
-#include <string>
-#include <cstring>
-
-using namespace std;
 
 char *openCVVersion() {
-    ostringstream os;
-    os << CV_VERSION;
-    return const_cast<char*>(os.str().c_str());
+    return CV_VERSION;
 }


deadprogram:
Hello @longlene thanks for the simplification. I changed the branch for your PR to `dev` as mentioned here: https://github.com/hybridgroup/gocv/blob/master/CONTRIBUTING.md

Thank you, now merging.

deadprogram:
Hello @longlene I merged the PR, but now I am seeing the following warning:

```
warning: ISO C++ forbids converting a string constant to ‘char*’ [-Wwrite-strings]
```

It *does* work as expected, but I am uncertain, perhaps it would fail under some other build config. What do you think?

longlene:
@deadprogram I build it without warning, maybe build parameter is not same.
Adding const declaration for the function openCVVersion can fix this warning.
Should I make a new pr on dev?

deadprogram:
Hi @longlene yes please, that would be great.

PR #23, opened 2017-11-05
fix the write-strings warning

longlene:
1. change the return temp string https://github.com/hybridgroup/gocv/pull/21
2. fix the write-strings warning,


deadprogram:
Thanks, @longlene for the update!

PR #32, opened 2017-11-20
Update env.sh for Homebrew

LAKuEN:
Fix to get the OpenCV path with "brew info".


deadprogram:
Hi @LAKuEN that seems like an excellent idea, much less brittle. Thanks!

PR #33, opened 2017-11-24
Feature/lut

lutskanu:
cv::LUT function binding


deadprogram:
Hi @lutskanu thank you very much for your PR.

Based on the location in OpenCV of the LUT function (in modules/core), it seems to me that this should be implemented in GoCV's core.go, core.cpp and core.h instead of in imgproc*.

Otherwise, looks great.

What do you think?

lutskanu:
Hi @deadprogram you're definitely right. I'll fix it soon.
Have you any suggestion what's wrong with the failed test? I don't see how this fail can be related to the new LUT binding.

deadprogram:
You need to rebase against `dev` as there is a change in the LDFLAGS, due to adding `features2d`:

https://github.com/hybridgroup/gocv/blob/dev/.travis.yml#L44

lutskanu:
Thank you, @deadprogram , I've fixed that!

deadprogram:
Thanks @lutskanu now merging.

PR #36, opened 2017-12-04
added LBPH facerecognizer bindings

andboson:


andboson:
need some hints :\

contrib/face.go
```diff
@@ -0,0 +1,147 @@
+package contrib
+
+/*
+#cgo LDFLAGS: -lopencv_face
```
deadprogram: In order to handle the correct build chain, all LDFLAGS should be set via env var. Remove this line, then add the `-lopencv_face` to the following 3 places:

- env.sh here https://github.com/hybridgroup/gocv/blob/master/env.sh#L12 and here https://github.com/hybridgroup/gocv/blob/master/env.sh#L16
- .travis.yml https://github.com/hybridgroup/gocv/blob/master/.travis.yml#L44
- appveyor.yml https://github.com/hybridgroup/gocv/blob/master/appveyor.yml#L29

deadprogram:
Now you are just missing a test for the new `SetNeighbors()` function, and all will pass the CI and the code coverage.

deadprogram:
Thanks for the contribution @andboson now merging.

PR #44, opened 2017-12-23
Features/colormap

221bytes:
The main feature of this branch is the addition of the colormap functions.
It fix a couple of typo mistake too (src instead of source).
Use the l2 norm for the WarpAffine function.


deadprogram:
This is great @221bytes thank you! :art: :world_map:

PR #45, opened 2017-12-24
Core/set values

221bytes:
Allow the edition of the Mat


deadprogram:
Looks great @221bytes thank you! Now merging.

PR #50, opened 2017-12-30
Add Mat ToBytes

jheidel:
This provides access to the Mat data pointer. It is useful for converting OpenCV Mat objects to other image types.

This mirrors the ability to use "tostring()" in the Python bindings. See:
https://stackoverflow.com/questions/5825173/pipe-raw-opencv-images-to-ffmpeg

I've also verified this works with real data by implementing an FFmpeg pipe in my own project.


deadprogram:
Looks good @jheidel thank you for the contribution. Now merging.

PR #54, opened 2018-01-10
add fileExt type

cnjack:

PR #55, opened 2018-01-10
fix err fileExt evoke panic

cnjack:


cnjack:
a better idea:(imgcodecs.go)
```
func IMEncode(fileExt FileExt, img Mat) (buf []byte, err error) {
	cfileExt := C.CString(string(fileExt))
	defer C.free(unsafe.Pointer(cfileExt))

	b := C.Image_IMEncode(cfileExt, img.Ptr())
	defer C.ByteArray_Release(b)
	if b.length == 0 {
		return nil, errors.New("encoder not found")
	}
	return toGoBytes(b), nil
}
```

deadprogram:
Hello @cnjack looks like the Windows version of the test you added failed:
https://ci.appveyor.com/project/deadprogram/gocv/build/216#L141

Perhaps it is more sensitive about the extension syntax `.jpg`?

cnjack:
`catch(...)`is not work in mingw?
I found something in [https://stackoverflow.com/questions/29271714/catch-across-shared-library-boundary-on-gcc-mingw](https://stackoverflow.com/questions/29271714/catch-across-shared-library-boundary-on-gcc-mingw)

deadprogram:
That figures... well options are to remove the try/catch entirely, or add some conditional platform specific code in the C++. What do you think?

deadprogram:
The Appveyor build appears to be using shared libs

`-DBUILD_SHARED_LIBS=ON` here:

https://github.com/hybridgroup/gocv/blob/master/appveyor_build_opencv.cmd#L13

cnjack:
maybe mingw can not use `try-catch`?

deadprogram:
I just commented here on this Go issue, seems like applies to this PR: https://github.com/golang/go/issues/13672#issuecomment-357169958

cnjack:
got

cnjack:
get it, may be it's ok in go1.10

deadprogram:
Hi @cnjack I do not think this will be fixed anytime soon in Go itself, from all the comments on those issues. However, thanks to replies from @ianlancetaylor a possible solution might be to simply not perform any operations that result in memory allocations taking place in the try/catch block. E.g. something like this:

```c++
struct ByteArray Image_IMEncode(const char* fileExt, Mat img) {
    std::vector<uchar> data;
    try{
        cv::imencode(fileExt, *img, data);
        return toByteArray(reinterpret_cast<const char*>(&data[0]), data.size());
    } catch(...) {
        // do not allocate memory here, just pass thru...
    }
    // the following line should only be reached in the case that an exception was encountered...
    return toByteArray(NULL, 0);
}
```

That might not be best way to structure the code, but it illustrates a technique that *might* work. Worth a try, I think.

cnjack:
I will try soon

cnjack:
sadly, it's not work

deadprogram:
Oh well. At least we tried that, thanks for making the effort.

cnjack:
is ok, wait the go fix the problem, the last pr #54 may avoid  problems such as these

PR #63, opened 2018-01-18
in the readme, explain compiling to a static library

### LINKED ISSUE #59: iteration times on Windows seem unbearable
I'm using Windows 10 with an i7-7700 CPU.  I followed the installation instructions from https://gocv.io/getting-started/windows/ (very well written, by the way--better than the opencv officials docs imo!) and everything did work.

Unfortunately, even toy 10-line programs that do something like "load an image from a file, do Canny edge detection, and display it in a window," are taking 10-15 seconds to compile and link.  Is it possible I've misconfigured something?  Is this unusual?  Or is this just "how things are" with the current state of cgo and large C++ libraries?

I don't want to be overly melodramatic, but as someone who's not already an OpenCV expert, this kind of iteration time is a deal breaker for experimenting and learning the library.

directive-cary:
Add sections in the readme that explain compiling gocv to a static library, to make it more easily consumable by projects that don't intend to modify gocv source.  Fixes #59 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,10 @@ The version program should output the following:
 
 You might want to copy the `env.sh` script into your own projects, to make it easier to setup these vars when building your own code.
 
+If you are not modifying gocv source, compile gocv to a static library, to significantly decrease your build times (`env.sh` must have been executed as described above):
+
+        go install gocv.io/x/gocv
+
 ### Other Linux installations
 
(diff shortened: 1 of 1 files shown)


deadprogram:
Thank you for the clarifications, and for looking into this. It does substantially improve the performance for anyone who wants to simply consume the library!

Now merging.

PR #65, opened 2018-01-18
Closing contours should also delete points

jondot:


deadprogram:
Looks good, thank you very much @jondot now merging.

Issue #69, opened 2018-02-01
Add cv::matchTemplate() bindings


littledot:
Hi,

I'm interested in adding `cv::matchTemplate()` bindings to GoCV. 
API: https://docs.opencv.org/3.2.0/df/dfb/group__imgproc__object.html#ga586ebfb0a7fb604b35a23d85391329be
Source: https://github.com/opencv/opencv/blob/master/modules/imgproc/src/templmatch.cpp#L1090

Are you OK with this?

deadprogram:
Seems like a good idea to me.

https://docs.opencv.org/3.4.0/df/dfb/group__imgproc__object.html#ga586ebfb0a7fb604b35a23d85391329be

deadprogram:
Thanks for the contribution @littledot I will close this issue once the next release goes out.

littledot:
Thanks for the quick turnaround. 

Just curious, how often do you cut a release? I plan to add a few more API bindings.

deadprogram:
We have been doing monthly releases so far.

deadprogram:
The latest release 0.9.0 from today now includes this code. Thank you!

PR #71, opened 2018-02-02
Add cv::matchTemplate() bindings.

littledot:
Tests are WIP.


littledot:
Fixes #69. This commit adds cv::matchTemplate() bindings.

littledot:
`toy.jpg` is cropped from `face.jpg`, so the confidence level should be extremely high. From my experiences in running the unit test, I get values like 0.99939847, so 0.95 should be a good threshold to fail.

deadprogram:
This looks great, @littledot thank you very much for working on it. Now merging.

PR #79, opened 2018-02-11
Support for non-brew installs of opencv on Darwin

kmuchmore:
If you need features not included in the brew install of OpenCV (like GStreamer) and build it yourself, then the env.sh script would not work on Mac. This change, after doing the check to get the brew path, checks if the path was found, if it doesn't return a path, assumes a non-brew install and sets the env variables to the default OpenCV install paths.


deadprogram:
Thanks for this @kmuchmore seems like a good idea. The homebrew recipe for OpenCV used to have some additional options for other software, but the last few version just install the "typical" ones.

PR #81, opened 2018-02-13
core: add Type() method to Mat

kmuchmore:
Noticed the NewMatFromBytes method took in a MatType object but there was no convenient way to save off that information from an existing Mat.


deadprogram:
Thanks for noticing that this was missing, and contributing. Now merging.

PR #83, opened 2018-02-14
Add needed environment variables for Linux

metal3d:
Some Linux distribution as Fedora need to add PKG_CONFIG_PATH and LD_LIBRARY_PATH to be sure to find libraries at compile time and execution time.
Because opencv is installed in `/usr/local`, we need to fix that paths.


deadprogram:
Hello @metal3d thank you for this contribution. I switched it to the `dev` branch as described here https://github.com/hybridgroup/gocv/blob/master/CONTRIBUTING.md#how-to-use-our-github-repository

Now merging, thanks again.

PR #84, opened 2018-02-14
Add distribution detection to deps rule

metal3d:
We are not all on Ubuntu. So we detect wich tool to use to install dependencies. Note that we must test "dnf" before "yum" because Fedora let "yum" alias to "dnf" with a warning message.

Dependencies are not all found from the Ubuntu list. But on Fedora, that ones seems to be ok. If you found other dependencies, please append them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,39 @@
 .ONESHELL:
 .PHONY: test deps download build clean
 
+
+# Package list for each well-known Linux distribution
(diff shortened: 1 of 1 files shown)


deadprogram:
@metal3d this is very useful, thank you very much.

In similar fashion to your other recent contribution, I have switched it to the `dev` branch as described here https://github.com/hybridgroup/gocv/blob/master/CONTRIBUTING.md#how-to-use-our-github-repository

Now merging this one also, thank you!.

metal3d:
Very happy to give a hand. Next time I'll do PR on dev branch, no problem ;)

Issue #47, opened 2017-12-28
Support void Mat::copyTo(OutputArray m, InputArray mask)?


hurkgu:

deadprogram:
Hi @hurkgu this will require a `CopyToWithMask()` function be added to core.

PR #85, opened 2018-02-15
Fixed unsigned int issue and added Mat.CopyToWithMask()

hobbs:
Two commits (dependent, because unsigned int issue blocked the test for CopyToWithMask()):
 * Updated type for Mat.GetUCharAt() and Mat.SetUCharAt() to reflect uint8 instead of int8
 * Added Mat.CopyToWithMask() per #47


deadprogram:
Thank you for the PR @hobbs now merging.

PR #94, opened 2018-02-23
add split subtract sum wrappers

### LINKED ISSUE #92: Explain or add something to get HSV pixel value
Hello,

Working on a little project, I need to change saturation of a `gocv.Mat`. At this time, I subsample ROI on on image and used `roi.ConvertTo(working, gocv.ColorBGRToHSV )`

Python returns a big list of tripplet (H, S, V) so we can change a channel, and rebuild Mat. But with gocv I can't find anything to do this.

I tried to understand how to make it with `working.GetIntAt(i, j)` but it's not working as expected (maybe I don't understand what is the Mat)

Rebulding image with `working.ConvertTo(out, MatTypeCV8U)` gives a stange grid, not a saturated image. So... Is there something that exists or gocv lakes of that kind of operation ?

berak:
... and tests.

also a C accessor function for the Mats array structure, since it is impossible to loop over it from cgo

#resolves #92


deadprogram:
Hi @berak thank you for your PR. I switched the branch to `dev` and resolved the resulting merge conflicts. Thanks again for the contributions!

berak:
ah, wrong branch, missed it, sorry. it's even [mentioned here](https://github.com/hybridgroup/gocv/blob/master/CONTRIBUTING.md)

PR #96, opened 2018-02-24
Dnn updates

berak:
opencv's dnn::readNetFrom* will throw cpp exceptions, if a prototxt or model was not found.

those have to be handled in the c++ code, so that users can check `net.Empty()` in their go code

### DIFF
--- a/cmd/caffe-classifier/main.go
+++ b/cmd/caffe-classifier/main.go
@@ -14,7 +14,7 @@
 //
 // How to run:
 //
-// 		go run ./cmd/caffe-classifier/main.go 0 [protofile] [modelfile] [descriptionsfile]
+//    go run ./cmd/caffe-classifier/main.go 0 [protofile] [modelfile] [descriptionsfile]
 //
 // +build example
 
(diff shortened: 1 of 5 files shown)


berak:
codecov complains about missing tests for darknet and torch ?

deadprogram:
Hello @berak a couple of things on this PR:

- Regarding exception handling, there was a PR that added but it was not merged due to issues with how Windows exception handling conflicts with Go. I'm not sure if the root cause was addressed in Go itself.
- Regarding DNN tests, take a look at the Travis build here: https://github.com/hybridgroup/gocv/blob/master/travis_build_opencv.sh#L70

and the Appveyor build here:
https://github.com/hybridgroup/gocv/blob/master/appveyor_build_opencv.cmd#L19-L20

We download the model files directly from their origins so as to not include the files in GoCV tiself.

berak:
@deadprogram , thanks for your insights !

i'll guess it's better to close it for now, and wait for a more refined solution.

> We download the model files directly from their origins

yes, i've seen it. however,  the yolo model is~300mb, so i was shy of adding it. also, there's no gocv code to extract the bounding boxes for this (or ssd) yet.

there's probably other regions in gocv, which are more important (and more fruitful to work on)   ;)

thanks again, for your most professional support here !

PR #97, opened 2018-02-24
divide, multiply

berak:


deadprogram:
Thank you @berak looks good to me! Now merging.

Issue #72, opened 2018-02-04
!empty() in function detectMultiScale


erickeller:
Hi there,

```
go version                                                                                                                                      
go version go1.9.3 linux/amd64                                                                                                                   
```

trying out your face recognition example:
https://gocv.io/writing-code/face-detect/

I bumped into the following error.

```
  what():  /tmp/opencv/opencv-3.4.0/modules/objdetect/src/cascadedetect.cpp:1698: error: (-215) !empty() in function detectMultiScale

SIGABRT: abort
PC=0x7f313f54f428 m=0 sigcode=18446744073709551610
signal arrived during cgo execution

goroutine 1 [syscall, locked to thread]:
runtime.cgocall(0x4ab1d0, 0xc420055df0, 0xc420055e18)
  /usr/local/go/src/runtime/cgocall.go:132 +0xe4 fp=0xc420055dc0 sp=0xc420055d80 pc=0x41c684
gocv.io/x/gocv._Cfunc_CascadeClassifier_DetectMultiScale(0xef5720, 0xf7da50, 0x0, 0x0)
  gocv.io/x/gocv/_obj/_cgo_gotypes.go:645 +0x57 fp=0xc420055df0 sp=0xc420055dc0 pc=0x4a72f7
gocv.io/x/gocv.(*CascadeClassifier).DetectMultiScale.func1(0xef5720, 0xf7da50, 0x0, 0x0)
  /home/users/me/workspace/go/src/gocv.io/x/gocv/objdetect.go:53 +0xb5 fp=0xc420055e38 sp=0xc420055df0 pc=0x4a8f05
gocv.io/x/gocv.(*CascadeClassifier).DetectMultiScale(0xc420055f30, 0xf7da50, 0x0, 0x0, 0x0)
  /home/users/me/workspace/go/src/gocv.io/x/gocv/objdetect.go:53 +0x63 fp=0xc420055eb0 sp=0xc420055e38 pc=0x4a85d3
main.main()
  /home/users/me/workspace/go/src/facedetect/main.go:49 +0x23c fp=0xc420055f80 sp=0xc420055eb0 pc=0x4a968c
runtime.main()
  /usr/local/go/src/runtime/proc.go:195 +0x226 fp=0xc420055fe0 sp=0xc420055f80 pc=0x443a86
runtime.goexit()
[...]

PR #98, opened 2018-02-25
cmd: check loading resources

berak:
... murphy's law.

let's try to avoid panic situations like #72 . simply, every read or load call must be checked.

weird: the caffe-classifier example fails gracefully, the tf-classifier still panics (given a wrong filename).


deadprogram:
Thank you for the PR @berak now merging.

PR #99, opened 2018-02-25
Add WarpPerspective and GetPerspectiveTransform

willdurand:
Hi,

I am learning OpenCV (like I started today) and chose your lib to write Go code. I am playing with image manipulation to be able to read content from pictures. I had to apply a perspective transformation, so I've added two new functions to the `gocv` API. I use these functions that way:

``` go
        // `corners` is also of type `[]image.Point` and contains the actual corners of the "thing" I am interested in the image.

	gridWidth := int(math.Max(widthTop, widthBottom))
	gridHeight := int(math.Max(heightLeft, heightRight))

	flatCorners := []image.Point{
		image.Pt(0, 0),
		image.Pt(gridWidth, 0),
		image.Pt(gridWidth, gridHeight),
		image.Pt(0, gridHeight),
	}

	m := gocv.GetPerspectiveTransform(corners, flatCorners)
	gocv.WarpPerspective(img, img, m, image.Pt(gridWidth, gridHeight))
```

~Because I am not entirely sure it is written correctly (did not find anything in the contributing file) and also because **I have to write tests** (but need to understand what to test exactly), **the PR is in WIP state**.~ Feel free to leave comments to help me/tell me what you'd expect.

Thanks,
William

### DIFF
--- a/core.go
+++ b/core.go
@@ -774,3 +774,18 @@ func toRectangles(ret C.Rects) []image.Rectangle {
 	}
 	return rects
 }
+
+func toCPoints(points []image.Point) C.struct_Points {
+	cPointSlice := make([]C.struct_Point, len(points))
+	for i, point := range points {
+		cPointSlice[i] = C.struct_Point{
+			x: C.int(point.X),
+			y: C.int(point.Y),
+		}
+	}
+
+	return C.struct_Points{
+		points: (*C.Point)(&cPointSlice[0]),
+		length: C.int(len(points)),
+	}
+}
--- a/core_test.go
+++ b/core_test.go
@@ -533,3 +533,16 @@ func TestScalar(t *testing.T) {
 		t.Error("Scalar has invalid value")
 	}
 }
+
+func TestToCPoints(t *testing.T) {
+	points := []image.Point{
+		image.Pt(0, 0),
+		image.Pt(1, 1),
+	}
+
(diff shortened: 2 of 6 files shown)


berak:
@willdurand , if you add new functions, you *must* add respective tests, too.

deadprogram:
Hi @willdurand thank you for working on this PR.

Regarding tests, so far we have concentrated on "smoke" tests. As long as your test exercises the new functions without failing, we consider it "tested". In other words, we are only testing the wrapper, not OpenCV itself.

Hopefully that makes sense. Thanks!

willdurand:
Ok, that makes sense, I'll add the tests then.

willdurand:
@deadprogram ok, tests have been added. I also introduced a helper function as I started by duplicating code... and I found we could reuse it elsewhere.

deadprogram:
Hi @willdurand I especially like PRs when they refactor and remove code! Now merging, thanks again.

SuporteSistemas:
Hi, I am trying to use the WarpPerspective function, however the compiler returns a `cannot use returnMat (type gocv.Mat) as type *gocv.Mat in argument to gocv.WarpPerspective` error for the second argument, but when I change it to a pointer variable I get a segmentation violation, how do I solve this? Have the function changed since this merge?

PR #106, opened 2018-02-26
core: Eigen Exp ExtractChannels

berak:


berak:
tests stall on EigenNonSymmetric, looks like an opencv bug to me ;)

deadprogram:
Hello @berak this PR has the same merge conflict from changes to `core_test.go`. Thanks again!

berak:
hehe, finally found the button, conflict solved ;)

berak:
conflicts resolved

berak:
apologies, i was simply going too fast here, since merging one pr will invoke conflicts on any other.

(learned something, i hope ..)

core.go
```diff
@@ -555,6 +556,63 @@ func CountNonZero(src Mat) int {
...
+	DctInverse                = DftInverse
+	DctRows                   = DftRows
+)
+
+// Dct Performs a forward or inverse discrete Cosine transform of 1D or 2D array.
+//
+// For further details, please see:
+// https://docs.opencv.org/3.4.0/d2/de8/group__core__array.html#ga85aad4d668c01fbd64825f589e3696d4
+//
+func Dct(src Mat, dst Mat, flags int) {
```
deadprogram: Hi @berak did you mean to make `flags` of type `DftFlags`? Seems like that was your intention here, which seems like a good thing.

berak: oh, ofc. 

mind, if i change Dft as well ?

deadprogram: Please do!

core.go
```diff
@@ -555,6 +557,82 @@ func CountNonZero(src Mat) int {
...
+// ConvertScaleAbs Scales, calculates absolute values, and converts the result to 8-bit.
+//
+// For further details, please see:
+// https://docs.opencv.org/3.4.0/d2/de8/group__core__array.html#ga3460e9c9f37b563ab9dd550c4d8c4e7d
+//
+func ConvertScaleAbs(src Mat, dst Mat, alpha float64, beta float64) {
+	C.Mat_ConvertScaleAbs(src.p, dst.p, C.double(alpha), C.double(beta))
+}
+
+// CopyMakeBorder Forms a border around an image.
```
willdurand: I think we should have real English sentences in doc blocks, so I would not add uppercase to the verb.

core.go
```diff
@@ -555,6 +557,82 @@ func CountNonZero(src Mat) int {
...
+		val4: C.double(value.A),
+	}
+
+	C.Mat_CopyMakeBorder(src.p, dst.p, C.int(top), C.int(bottom), C.int(left), C.int(right), C.int(borderType), c_value)
+}
+
+type DftFlags int
+
+const (
+	DftForward DftFlags = 0
```
willdurand: We should have doc for the flags, that would be very useful.

core.go
```diff
@@ -555,6 +557,82 @@ func CountNonZero(src Mat) int {
...
+}
+
+// CopyMakeBorder Forms a border around an image.
+//
+// For further details, please see:
+// https://docs.opencv.org/3.4.0/d2/de8/group__core__array.html#ga2ac1049c2c3dd25c2b41bffe17658a36
+//
+func CopyMakeBorder(src Mat, dst Mat, top int, bottom int, left int, right int, borderType int, value color.RGBA) {
+
+	c_value := C.struct_Scalar{
```
willdurand: Shouldn't this be `cValue`? I am asking because in some other places, that was the case.

core.go
berak: indeed.

core.go
berak: right.

core.go
berak: an implementation detail, but ok....

deadprogram:
Thank you, now merging.

berak:
thanks ;)

PR #107, opened 2018-02-27
Add DrawContours()

willdurand:
This PR adds the `DrawContours()` function.

The test case creates the following image:

before

It finds the contours of the square and draws them with a larger border:

after

---

~~This PR works only when disabling CGO checks (`GODEBUG=cgocheck=0`) because I
am passing a struct of structs (slice of slices) and it breaks one of the CGO
"rules" I guess.~~

~~I am not really sure to know how to fix this to be honest, maybe we would need
a different C structure to handle multidimensional arrays of `cv::Point` and
being able to pass such data structures from Go to C. In `FindContours()`, we
have done the opposite and it seems to work, so there must be a way, but I just
don't know what to do here right now.~~

### DIFF
--- a/imgproc.cpp
+++ b/imgproc.cpp
@@ -270,3 +270,20 @@ Mat GetPerspectiveTransform(Contour src, Contour dst) {
 
   return new cv::Mat(cv::getPerspectiveTransform(src_pts, dst_pts));
 }
+
+void DrawContours(Mat src, Contours contours, int contourIdx, Scalar color, int thickness) {
+  std::vector<std::vector<cv::Point> > cntrs;
+  for (size_t i = 0; i < contours.length; i++) {
+    Contour contour = contours.contours[i];
+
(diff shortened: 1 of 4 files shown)


willdurand:
I have made progress here. It works as intended and I changed the way the 2d array is constructed. It should be fine although I did not check for memory leaks (yet).

deadprogram:
Hi @willdurand that is great that you are almost there. The Travis build indicates that there is an error in the C++ code here that needs to be corrected:

https://travis-ci.org/hybridgroup/gocv/builds/347507779#L862

willdurand:
mmh indeed, it compiles just well on my laptop, so I assumed tests were failing because coverage was not OK...  I'll have a look soon, thanks.

willdurand:
@deadprogram tests added.

deadprogram:
I was really needing this function @willdurand but had not gotten around to working on it. Thank you! Now merging.

PR #113, opened 2018-03-04
Tracking

berak:
an attempt to wrap the tracking module from opencv_contrib

W I P , rfc.

this is making heavy use of c++ / go interfaces.

Tracker_Init() and Tracker_Update() only need to be implemented once (in c++), and can be reused for any go implentation of the Tracker interface, [saving a ton of boilerplate code](https://www.brainyquote.com/quotes/ken_thompson_254858)

if we'd use interfaces more often, properly, e.g, there'd be no need to do this [again](https://github.com/hybridgroup/gocv/blob/master/features2d.go#L52-L58) and [again](https://github.com/hybridgroup/gocv/blob/master/features2d.go#L133-L139) , and also have [all this boilerplate repetition here](https://github.com/hybridgroup/gocv/blob/master/features2d.go#L133-L139) and [here](https://github.com/hybridgroup/gocv/blob/master/features2d.cpp#L72-L98) and so on ..

@willdurand , may i challenge you / pick your mind on this idea, too ?

(it still needs a wrapper function on the go side, with the correct implementation type as self, to satisfy the go interface, but all we need to do there is to extract the C.Tracker from the go type, and pass it on to the shared c++ wrapper)

also tests can be rigorously simplified, if all we need to do is pass a different implementation instance to the same base test

some things did not make it:

* GOTURN:  needs a XXXmb pre-trained dnn, can't be tested without (throws exception)
<strike> * CSRT: only available in 3.4.1</strike> .. i'll just add it
* MultiTracker: i've made an attempt, but there are problems. i'll leave the code in comments (for now), maybe someone has an idea.
    a: i see no proper way to add new tracker instances to it. the way i tried, worked on my win box, but throws an exception on travis. it's also somewhat unclear, how & when to Close() the added trackers
   b: i'm unable to release a `C.struct_RECTS` in cgo, (can't access `Rects_Close` somehow)
   c: `gocv.toRectangles(ret)` would be very useful here, but it's private in gocv
all in all, opencv's MultiTracker is just a shallow wrapper around `vector<Ptr<Tracker>>`, maybe we're better off, building a mock class in pure go ?

### DIFF
--- a/cmd/tracking/main.go
+++ b/cmd/tracking/main.go
@@ -1,0 +1,105 @@
+// What it does:
+//
+// This example uses one of the Tracker classes from opencv_contrib to track a region of interest (e.g. a face)
+// and draws a rectangle around it, before displaying it within a Window.
+//
+// in this example, users have to select an initial roi with the mouse, and press enter, to start the tracking
+// (but the roi could also come from e.g. a previous cascade based detection)
+//
+// also see https://docs.opencv.org/master/d2/d0a/tutorial_introduction_to_tracker.html for a tutorial
+//
+// How to run:
+//
+// tracking [camera ID]
+//
+// 		go run ./cmd/tracking/main.go 0
+//
+// +build example
+
+package main
+
+import (
+	"fmt"
+	"image/color"
+	"os"
+	"strconv"
+
+	"gocv.io/x/gocv"
+	"gocv.io/x/gocv/contrib"
+)
+
+func main() {
+	if len(os.Args) < 2 {
+		fmt.Println("How to run:\n\ttracking [camera ID]")
+		return
+	}
+
+	// parse args
+	deviceID, _ := strconv.Atoi(os.Args[1])
+
+	// open webcam
+	webcam, err := gocv.VideoCaptureDevice(int(deviceID))
+	if err != nil {
+		fmt.Printf("error opening video capture device: %v\n", deviceID)
+		return
+	}
+	defer webcam.Close()
+
+	// open display window
+	window := gocv.NewWindow("Tracking")
+	defer window.Close()
+
+	// create a tracker instance
+	// (one of Mil,Kcf,Tld,MedianFlow,Boosting or Mosse)
+	tracker := contrib.NewTrackerMosse()
+	defer tracker.Close()
+
+	// prepare image matrix
+	img := gocv.NewMat()
+	defer img.Close()
+
+	// read an initial image
+	if ok := webcam.Read(img); !ok {
+		fmt.Printf("cannot read device %d\n", deviceID)
+		return
+	}
+
+	// let the user mark a ROI to track
+	rect := gocv.SelectROI("Tracking", img)
+	if rect.Max.X == 0 {
+		fmt.Printf("user cancelled roi selection\n")
+		return
+	}
+
+	// initialize the tracker with the image & the selected roi
+	init := tracker.Init(img, rect)
+	if !init {
+		fmt.Printf("Could not initialize the Tracker")
+		return
+	}
+
+	// color for the rect to draw
+	blue := color.RGBA{0, 0, 255, 0}
+	fmt.Printf("start reading camera device: %v\n", deviceID)
+	for {
+		if ok := webcam.Read(img); !ok {
+			fmt.Printf("cannot read device %d\n", deviceID)
+			return
+		}
+		if img.Empty() {
+			continue
+		}
+
+		// update the roi
+		rect, _ := tracker.Update(img)
+
+		// draw it.
+		gocv.Rectangle(img, rect, blue, 3)
+
+		// show the image in the window, and wait 10 millisecond
+		window.IMShow(img)
+		if window.WaitKey(10) >= 0 {
+			break
+		}
+	}
+}
--- a/cmd/tracking/main.go
+++ b/cmd/tracking/main.go
@@ -50,8 +50,8 @@ func main() {
 	defer window.Close()
 
 	// create a tracker instance
-	// (one of Mil,Kcf,Tld,MedianFlow,Boosting or Mosse)
-	tracker := contrib.NewTrackerMosse()
+	// (one of MIL, KCF, TLD, MedianFlow, Boosting, MOSSE or CSRT)
+	tracker := contrib.NewTrackerMOSSE()
 	defer tracker.Close()
 
 	// prepare image matrix
--- a/contrib/tracking.cpp
+++ b/contrib/tracking.cpp
@@ -3,34 +3,29 @@
 
 
 bool Tracker_Init(Tracker self, Mat image, Rect boundingBox) {
-    cv::Rect2d c_boundingBox(boundingBox.x, boundingBox.y, boundingBox.width, boundingBox.height);
+    cv::Rect2d bb(boundingBox.x, boundingBox.y, boundingBox.width, boundingBox.height);
 
-    bool ret = (*self)->init(*image, c_boundingBox);
+    bool ret = (*self)->init(*image, bb);
     return ret;
 }
-
 
 bool Tracker_Update(Tracker self, Mat image, Rect *boundingBox) {
-    cv::Rect2d cBox;
-    bool ret = (*self)->update(*image, cBox);
-    boundingBox->x = int(cBox.x);
-    boundingBox->y = int(cBox.y);
-    boundingBox->width = int(cBox.width);
-    boundingBox->height = int(cBox.height);
+    cv::Rect2d bb;
+    bool ret = (*self)->update(*image, bb);
+    boundingBox->x = int(bb.x);
+    boundingBox->y = int(bb.y);
+    boundingBox->width = int(bb.width);
+    boundingBox->height = int(bb.height);
     return ret;
 }
 
-
-
-
-TrackerMil TrackerMil_Create() {
+TrackerMIL TrackerMIL_Create() {
     return new cv::Ptr<cv::TrackerMIL>(cv::TrackerMIL::create());
 }
 
-void TrackerMil_Close(TrackerMil self){
+void TrackerMIL_Close(TrackerMIL self){
     delete self;
 }
-
 
 TrackerBoosting TrackerBoosting_Create() {
     return new cv::Ptr<cv::TrackerBoosting>(cv::TrackerBoosting::create());
@@ -40,7 +35,6 @@ void TrackerBoosting_Close(TrackerBoosting self){
 	delete self;
 }
 
-
 TrackerMedianFlow TrackerMedianFlow_Create() {
     return new cv::Ptr<cv::TrackerMedianFlow>(cv::TrackerMedianFlow::create());
 }
@@ -49,39 +43,34 @@ void TrackerMedianFlow_Close(TrackerMedianFlow self){
 	delete self;
 }
 
-
-TrackerTld TrackerTld_Create() {
+TrackerTLD TrackerTLD_Create() {
     return new cv::Ptr<cv::TrackerTLD>(cv::TrackerTLD::create());
 }
 
-void TrackerTld_Close(TrackerTld self){
+void TrackerTLD_Close(TrackerTLD self){
     delete self;
 }
 
-
-TrackerKcf TrackerKcf_Create() {
+TrackerKCF TrackerKCF_Create() {
     return new cv::Ptr<cv::TrackerKCF>(cv::TrackerKCF::create());
 }
 
-void TrackerKcf_Close(TrackerKcf self){
+void TrackerKCF_Close(TrackerKCF self){
     delete self;
 }
 
-
-TrackerMosse TrackerMosse_Create() {
+TrackerMOSSE TrackerMOSSE_Create() {
     return new cv::Ptr<cv::TrackerMOSSE>(cv::TrackerMOSSE::create());
 }
 
-void TrackerMosse_Close(TrackerMosse self){
+void TrackerMOSSE_Close(TrackerMOSSE self){
     delete self;
 }
 
-
-
-TrackerCsrt TrackerCsrt_Create() {
+TrackerCSRT TrackerCSRT_Create() {
     return new cv::Ptr<cv::TrackerCSRT>(cv::TrackerCSRT::create());
 }
 
-void TrackerCsrt_Close(TrackerCsrt self){
+void TrackerCSRT_Close(TrackerCSRT self){
     delete self;
 }
--- a/contrib/tracking.cpp
+++ b/contrib/tracking.cpp
@@ -1,0 +1,87 @@
+#include "tracking.h"
+#include <opencv2/opencv.hpp>
+
+
+bool Tracker_Init(Tracker self, Mat image, Rect boundingBox) {
+    cv::Rect2d c_boundingBox(boundingBox.x, boundingBox.y, boundingBox.width, boundingBox.height);
+
+    bool ret = (*self)->init(*image, c_boundingBox);
+    return ret;
+}
+
+
+bool Tracker_Update(Tracker self, Mat image, Rect *boundingBox) {
+    cv::Rect2d cBox;
+    bool ret = (*self)->update(*image, cBox);
+    boundingBox->x = int(cBox.x);
+    boundingBox->y = int(cBox.y);
+    boundingBox->width = int(cBox.width);
+    boundingBox->height = int(cBox.height);
+    return ret;
+}
+
+
+
+
+TrackerMil TrackerMil_Create() {
+    return new cv::Ptr<cv::TrackerMIL>(cv::TrackerMIL::create());
+}
+
+void TrackerMil_Close(TrackerMil self){
+    delete self;
+}
+
+
+TrackerBoosting TrackerBoosting_Create() {
+    return new cv::Ptr<cv::TrackerBoosting>(cv::TrackerBoosting::create());
+}
+
+void TrackerBoosting_Close(TrackerBoosting self){
+	delete self;
+}
+
+
+TrackerMedianFlow TrackerMedianFlow_Create() {
+    return new cv::Ptr<cv::TrackerMedianFlow>(cv::TrackerMedianFlow::create());
+}
+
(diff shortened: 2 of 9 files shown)


berak:
i don't understand the appveyor problem. can anyone help ?

deadprogram:
This is a very interesting PR, I will look at it more closely in the next couple of days...

willdurand:
The code you wrote is easy to read/follow and if we can DRY, let's do it more often. It seems the single Tracker feature is implemented, so maybe you could cleanup that part, and we talk about the Multi Tracker in a separate PR?

berak:
@willdurand , thanks for the kind review.

for now, i think, we're better off without the multitracker (in the library)

the c++ version had a couple of warts, eg, if a single tracker loses its target, you never know which one, and you'd have to throw the whole thing away, and reconstruct it from scratch.

i'll try to make a go based one, and maybe it could be a simple cmd sample ?

willdurand:
Alright, ping me when you want a review.

> the c++ version had a couple of warts, eg, if a single tracker loses its target, you never know which one, and you'd have to throw the whole thing away, and reconstruct it from scratch.

Does that make sense from an opencv point of view? Or does opencv handle things differently and we should try to do that too in Go?

willdurand:
@berak can I review this PR by chance?

contrib/tracking.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+
+bool Tracker_Init(Tracker self, Mat image, Rect boundingBox) {
+    cv::Rect2d c_boundingBox(boundingBox.x, boundingBox.y, boundingBox.width, boundingBox.height);
+
+    bool ret = (*self)->init(*image, c_boundingBox);
+    return ret;
+}
+
+
```
willdurand: This comment applies to the whole file: I think we should only have one blank line between the functions.

contrib/tracking.cpp
```diff
@@ -0,0 +1,87 @@
...
+TrackerMedianFlow TrackerMedianFlow_Create() {
+    return new cv::Ptr<cv::TrackerMedianFlow>(cv::TrackerMedianFlow::create());
+}
+
+void TrackerMedianFlow_Close(TrackerMedianFlow self){
+	delete self;
+}
+
+
+TrackerTld TrackerTld_Create() {
```
willdurand: Could this be `TrackerTLD` since the `cv` namespace or class is `TrackerTLD`?

contrib/tracking.cpp
```diff
@@ -0,0 +1,87 @@
...
+    return new cv::Ptr<cv::TrackerTLD>(cv::TrackerTLD::create());
+}
+
+void TrackerTld_Close(TrackerTld self){
+    delete self;
+}
+
+
+TrackerKcf TrackerKcf_Create() {
```
willdurand: Same question for `TrackerKCF`

contrib/tracking.cpp
```diff
@@ -0,0 +1,87 @@
...
+    return new cv::Ptr<cv::TrackerKCF>(cv::TrackerKCF::create());
+}
+
+void TrackerKcf_Close(TrackerKcf self){
+    delete self;
+}
+
+
+TrackerMosse TrackerMosse_Create() {
```
willdurand: Same here for `TrackerMOSSE`

contrib/tracking.cpp
```diff
@@ -0,0 +1,87 @@
...
+    return new cv::Ptr<cv::TrackerMOSSE>(cv::TrackerMOSSE::create());
+}
+
+void TrackerMosse_Close(TrackerMosse self){
+    delete self;
+}
+
+
+
+TrackerCsrt TrackerCsrt_Create() {
```
willdurand: Same here for `TrackerCSRT`

contrib/tracking.go
```diff
@@ -0,0 +1,256 @@
...
+import "C"
+import (
+	"image"
+
+	"gocv.io/x/gocv"
+)
+
+// Tracker - the base interface for object tracking
+//
+// s.a: https://docs.opencv.org/master/d0/d0a/classcv_1_1Tracker.html
```
willdurand: what does "s.a" stand for actually?

contrib/tracking.go
```diff
@@ -0,0 +1,256 @@
...
+#include "tracking.h"
+*/
+import "C"
+import (
+	"image"
+
+	"gocv.io/x/gocv"
+)
+
+// Tracker - the base interface for object tracking
```
willdurand: > tracker is the base interface for object tracking.

contrib/tracking.go
```diff
@@ -0,0 +1,256 @@
...
+// s.a: https://docs.opencv.org/master/d0/d0a/classcv_1_1Tracker.html
+type Tracker interface {
+	// Init initializes the tracker with a known bounding box that surrounded the target
+	// note: this can only be called once. if you loose the object, you have to Close() the instance,
+	// create a new one, and call Init on it again
+	//
+	// https://docs.opencv.org/master/d0/d0a/classcv_1_1Tracker.html#a4d285747589b1bdd16d2e4f00c3255dc
+	Init(image gocv.Mat, boundingBox image.Rectangle) bool
+
+	// Update the tracker, return new bounding box and a bool, if it lost it.
```
willdurand: > Update updates the tracker, returns a new bounding box and a boolean determining whether the tracker lost the target.

(correct me if this is wrong but that's what I understand)

contrib/tracking.go
```diff
@@ -0,0 +1,256 @@
...
+//
+// available implementations in gocv are:
+//
+// TrackerMil;
+// TrackerBoosting;
+// TrackerMedianFlow;
+// TrackerTld;
+// TrackerKcf;
+// TrackerMosse;
+// TrackerCsrt;
```
willdurand: I dont think we really need this anymore

contrib/tracking.go
```diff
@@ -0,0 +1,256 @@
...
+//
+
+// (private) implementation helpers for init and update (saves some boilerplate code)
```
willdurand: I don't think we need to prefix with `(private)` as Go functions are private when the first char is lowercase.

willdurand: also, I'd say "(saves some boilerplate code)" does not carry a lot of useful information

contrib/tracking.go
```diff
@@ -0,0 +1,256 @@
...
+func tracker_Init(trk C.Tracker, img gocv.Mat, boundingBox image.Rectangle) bool {
+	cBox := C.struct_Rect{
+		x:      C.int(boundingBox.Min.X),
+		y:      C.int(boundingBox.Min.Y),
+		width:  C.int(boundingBox.Size().X),
```
willdurand: How different `boundingBox.Size().X` and `boundingbox.Max.X` are?

contrib/tracking.go
```diff
@@ -0,0 +1,256 @@
...
+
+	ret := C.Tracker_Update(trk, C.Mat(img.Ptr()), &cBox)
+
+	rect := image.Rect(int(cBox.x), int(cBox.y), int(cBox.x+cBox.width), int(cBox.y+cBox.height))
+	return rect, bool(ret)
+}
+
+//
+// Mil
+// The MIL algorithm trains a classifier in an online manner to separate the object from the background.
```
willdurand: It should starts with:

> TrackerMil implements the MIL algorithm, which trains a classifier [...]

It is important because here is what I currently get on godoc:

screen shot 2018-03-12 at 12 13 27

I care a lot about documentation because this is the only thing this project offers me (as a user) to use it.

willdurand: This comment applies to all trackers then.

contrib/tracking.go
```diff
@@ -0,0 +1,256 @@
...
+// Multiple Instance Learning avoids the drift problem for a robust tracking.
+//
+// s.a: https://docs.opencv.org/master/d0/d26/classcv_1_1TrackerMIL.html
+type TrackerMil struct {
+	p C.TrackerMil
+}
+
+func NewTrackerMil() TrackerMil {
```
willdurand: Missing godoc here.

contrib/tracking.go
```diff
@@ -0,0 +1,256 @@
...
+	return TrackerMil{p: C.TrackerMil_Create()}
+}
+
+func (self *TrackerMil) Close() error {
```
willdurand: Missing godoc here.

contrib/tracking.go
```diff
@@ -0,0 +1,256 @@
...
+	C.TrackerMil_Close(self.p)
+	self.p = nil
+	return nil
+}
+
+func (self TrackerMil) Init(img gocv.Mat, boundingBox image.Rectangle) bool {
```
willdurand: Missing godoc here.

contrib/tracking.go
```diff
@@ -0,0 +1,256 @@
...
+	return tracker_Init(C.Tracker(self.p), img, boundingBox)
+}
+
+func (self TrackerMil) Update(img gocv.Mat) (image.Rectangle, bool) {
```
willdurand: Missing godoc here.

contrib/tracking.h
```diff
@@ -0,0 +1,70 @@
+
+#ifndef _OPENCV3_MY_H_
```
willdurand: Why is this `MY` here and not something `TRACKING`?

contrib/tracking_test.go
```diff
@@ -0,0 +1,65 @@
...
+func BaseTestTracker(t *testing.T, tracker Tracker, name string) {
+	if tracker == nil {
+		t.Error("TestTracker " + name + " should not be nil")
+	}
+	img := gocv.IMRead("../images/face.jpg", 1)
+	if img.Empty() {
+		t.Error("TestTracker " + name + " input img failed to load")
+	}
+	rect := image.Rect(250, 150, 250+200, 150+250)
+	init := tracker.Init(img, rect)
```
willdurand: maybe we should assert on the bounding box values to make sure we do the math correctly.

contrib/tracking_test.go
```diff
@@ -0,0 +1,65 @@
...
+	if !ok {
+		t.Error("TestTracker " + name + " lost object in Update")
+	}
+	img.Close()
+}
+
+func TestTrackerMil(t *testing.T) {
+	tracker := NewTrackerMil()
+	BaseTestTracker(t, tracker, "Mil")
+	tracker.Close()
```
willdurand: I guess `defer tracker.Close()` would work, right?

contrib/tracking_test.go
```diff
@@ -0,0 +1,65 @@
...
+	tracker := NewTrackerMosse()
+	BaseTestTracker(t, tracker, "Mosse")
+	tracker.Close()
+}
+
+func TestTrackerCsrt(t *testing.T) {
+	tracker := NewTrackerCsrt()
+	BaseTestTracker(t, tracker, "Csrt")
+	tracker.Close()
+}
```
willdurand: Instead of creating many test functions, we could use a [table-driven test](https://github.com/golang/go/wiki/TableDrivenTests), WDYT?

willdurand: I really like the DRY approach BTW

contrib/tracking.go
berak: > what I currently get on godoc

oh, haven't tried at all, but sure, it should make sense in the godoc context (vs humans reading src code)

contrib/tracking.cpp
berak: camel or not ?

(they were all like it is in opencv, origially)

contrib/tracking.go
berak: see also

contrib/tracking.go
berak: yea, better.

contrib/tracking.go
berak: Max.X = Min.X + width

(opencv's rects are [x,y,w,h] go's are [x0,y0, x1,y1])

contrib/tracking.h
berak: k,k.

contrib/tracking_test.go
berak: good idea, a table of functions was a bit over my head, though. will have to look it up.

contrib/tracking.go
berak: hmm, isn't that only useless repetition ?

(those are all boring implementations of the same interface)

(but i'll have to look at the godoc output, again, let's see)

contrib/tracking_test.go
berak: the img is 640x480

contrib/tracking.go
willdurand: ah no it makes sense, thank you! Google did not help me..

contrib/tracking.go
willdurand: oh ok, thanks for the explanation

contrib/tracking_test.go
willdurand: @berak try something like this:

```go
func TestTrackers(t *testing.T) {
	tab := []struct {
		name    string
		tracker Tracker
	}{
		{"Mil", NewTrackerMil()},
		{"Boosting", NewTrackerBoosting()},
	}

	for _, tt := range tab {
		if tt.tracker == nil {
			t.Error("TestTracker " + tt.name + " should not be nil")
		}

		img := gocv.IMRead("../images/face.jpg", 1)
		defer img.Close()

		if img.Empty() {
			t.Error("TestTracker " + tt.name + " input img failed to load")
		}

		rect := image.Rect(250, 150, 250+200, 150+250)

		if init := tt.tracker.Init(img, rect); !init {
			t.Error("TestTracker " + tt.name + " failed in Init")
		}

		_, ok := tt.tracker.Update(img)
		if !ok {
			t.Error("TestTracker " + tt.name + " lost object in Update")
		}
	}
}
```

contrib/tracking.go
berak: a single   "see:"  would be better, i guess

contrib/tracking.cpp
berak: but yes, happy to change it back !

contrib/tracking_test.go
berak: hehe, that's what i'm doing now ;)

(and it revealed the problem with Close, since all we have is a Tracker now)

contrib/tracking.go
willdurand: We often use "See also:" too

contrib/tracking.cpp
willdurand: I dont know, I like the Go way (`TrackerTLD`, etc.) but it is up to you. What matters the most is to be consistent.

berak:
* changed class names back to opencv's upper case
* table based test
* misc. doc fixes

contrib/tracking.go
willdurand: why stylesheet? I use `godoc -http 127.0.0.1:6060` only.

contrib/tracking.go
willdurand: Yes and no, it is a bit repetitive but it is useful to have documentation everywhere so that when you're looking for something, you dont have to look somewhere else first, even if that "somewhere" looks similar to where you were initially.

cmd/tracking/main.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	defer webcam.Close()
+
+	// open display window
+	window := gocv.NewWindow("Tracking")
+	defer window.Close()
+
+	// create a tracker instance
+	// (one of Mil,Kcf,Tld,MedianFlow,Boosting or Mosse)
+	tracker := contrib.NewTrackerMosse()
```
willdurand: so I think you meant `NewTrackerMOSSE` now

berak: harrr....

contrib/tracking_test.go
```diff
@@ -0,0 +1,51 @@
...
+		{"TLD", NewTrackerTLD()},
+		{"KCF", NewTrackerKCF()},
+		{"MOSSE", NewTrackerMOSSE()},
+		{"CSRT", NewTrackerCSRT()},
+	}
+
+	for _, test := range tab {
+		defer test.tracker.Close()
+
+		BaseTestTracker(t, test.tracker, test.name)
```
willdurand: I would actually move the code of `BaseTestTracker` here. I'd do that because it would make more sense to have all the code in the test case for me but also because the name "BaseTestTracker" does not really mean anything. If you want to keep the function outside for some reason, name it like `assertTracker` for example.

willdurand:
More a general comment, it is quite important to respect the way godoc is usually written, it should always start with the name of the thing it describes (see https://blog.golang.org/godoc-documenting-go-code).

Example:

> TrackerMIL provides the MIL algorithm that trains a classifier in an online manner to separate the object from the background.

> TrackerBoosting provides a real-time object tracker based on a novel on-line version of the AdaBoost algorithm.

Also, you really don't need [blank comment lines](https://github.com/hybridgroup/gocv/pull/113/files#diff-20e54db4a051cdc985f350afc0d03b33R19).

Good job BTW, it is lovely to have this tracking API 😍

berak:
codebase if full of that https://github.com/hybridgroup/gocv/blob/master/core.go#L89

berak:
trackpad failure, sorry

willdurand:
> codebase if full of that https://github.com/hybridgroup/gocv/blob/master/core.go#L89

right, that's why I did not bother commenting on each line 😛

deadprogram:
@berak and @willdurand this PR is looking great. Thank you both. Seems like it now ready for merge, correct?

berak:
no objection ;)

willdurand:
I am fine even though I think the godoc could get some more love :)

berak:
@willdurand i think, i found a good way to resolve the remaining issues:

    keep the implementation classes for the Tracker interface package private.

this solves some problems:
* we only need the "constructors" in the API
* no need to document the implementation classes, avoid repeating /clutter in godoc

willdurand:
Yup, looks good to me.

deadprogram:
Great work! Now merging...

PR #115, opened 2018-03-04
core: added Pow() method

garciademarina:


deadprogram:
Thank you for the PR @garciademarina now merging.

PR #119, opened 2018-03-06
dnn: add a ssd sample and a GetBlobChannel helper

berak:
* GetBlobChannel : used to retrieve channels from a 4d blob. Mat.Ptr() is already taken, and this is quite specific to the dnn code, so i add a helper there (+ a test, ofc.).
* ssd sample: a demo how to use ssd networks, this one specifically does face detection. small model (5mb only !) , and much faster than the cascades !
(but the extraction part would be exactly the same for other ssd models, or even YOLO)


cmd/dnn-ssd/main.go
```diff
@@ -0,0 +1,140 @@
+// What it does:
+//
+// This example shows, how to use pretrained SSD (Single Shot Detection) detection networks in gocv.
+// here, we detect human faces from the camera, but the setup is similar for any other kind of object detection.
```
willdurand: s/here/Here

cmd/dnn-ssd/main.go
```diff
@@ -0,0 +1,140 @@
...
+	"fmt"
+	"image"
+	"image/color"
+	"os"
+	"strconv"
+
+	"gocv.io/x/gocv"
+)
+
+func min(a, b float32) float32 {
```
willdurand: Is it because you use `float32` that you cant use [`math.Max`](https://golang.org/pkg/math/#Max)?

cmd/dnn-ssd/main.go
```diff
@@ -0,0 +1,140 @@
...
+		// convert image Mat to 96x128 blob that the detector can analyze
+		blob := gocv.BlobFromImage(img, 1.0, image.Pt(128, 96), gocv.NewScalar(104.0, 177.0, 123.0, 0), false, false)
+
+		// feed the blob into the classifier
+		net.SetInput(blob, "data")
+
+		// run a forward pass thru the network
+		detBlob := net.Forward("detection_out")
+
+		// extract the detections (nfaces x 7features)
```
willdurand: s/7features/7 features

cmd/dnn-ssd/main.go
```diff
@@ -0,0 +1,140 @@
...
+			right = min(max(0, right), W-1)
+			bottom = min(max(0, bottom), H-1)
+			top = min(max(0, top), H-1)
+
+			// draw it
+			rect := image.Rect(int(left), int(top), int(right), int(bottom))
+			gocv.Rectangle(img, rect, green, 3)
+		}
+
+		blob.Close()
```
willdurand: shouldn't we `defer` the `Close()` calls earlier?

cmd/dnn-ssd/main.go
berak: yes, indeed. go offers only Min, Max for float64

cmd/dnn-ssd/main.go
berak: ok.

cmd/dnn-ssd/main.go
berak: needs better explanation, anyway

cmd/dnn-ssd/main.go
willdurand: Should we add a "utils" lib in gocv? These `min`/`max` functions are pretty useful.

cmd/dnn-ssd/main.go
willdurand: 😄

cmd/dnn-ssd/main.go
berak: maybe even join that with the test_helper.go, which has round()

willdurand: right, although I was thinking about exposing these functions in userland

berak: more like: rename helpers_test.go to utility.go, and make it public

willdurand: on the other hand, we have so much under the `gocv` "package" already that maybe we should thing about a sub-package and also we need to be smart because we likely need these functions for `int`, `float32`, etc.

Maybe we could gather the functions we use all the time when working with opencv that dont exist natively in Go, and create a sub-package: `gocv.math` with `MinInt`, `MinInt64`, `MinFloat32`, etc. WDYT? Just thinking it out loud..

berak: yea, the types are a problem, and that would solve it, imho.

berak:
added another helper func (GetBlobSize) and a pose detection demo



also found a small bug in Net_GetBlobsFromImage (crop value was not copied to the c++ code)

deadprogram:
This is so incredibly cool! Thank you for this contribution, now merging.

dnn.go
```diff
@@ -109,3 +109,18 @@ func BlobFromImage(img Mat, scaleFactor float64, size image.Point, mean Scalar,
...
+// GetBlobChannel extracts a single (2d)channel from a 4 dimensional blob structure
+// (this might e.g. contain the results of a SSD or YOLO detection,
+//  a bones structure from pose detection, or a color plane from Colorization)
+//
+func GetBlobChannel(blob Mat, imgidx int, chnidx int) Mat {
+	return Mat{p: C.Net_GetBlobChannel(blob.p, C.int(imgidx), C.int(chnidx))}
+}
+
+// GetBlobSize retrieves the 4 dimensional size information in (N,C,H,W) order
+//
```
willdurand: This was not really needed, and we usually put a period at the end of the line because we write sentences :)

berak: ok, your point.

you mean, the function was noot needed ? which functionality does it duplicate ?

willdurand: ah sorry for the brevity of my comment, I meant the extra commented line was not needed.

Issue #118, opened 2018-03-06
Subtract is defined twice in core.h


ibadami:
Function `Mat_Subtract` is defined twice in core.h at line 231 and line 235.

deadprogram:
Code for this fix has been merged into teh `dev` branch.

deadprogram:
This has been addressed as part of v0.11.0 thank you. Now closing.

PR #120, opened 2018-03-06
core: remove duplicate Mat_Subtract decraration

berak:
#118

### DIFF
--- a/core.h
+++ b/core.h
@@ -229,7 +229,6 @@ void Mat_MeanStdDev(Mat src, Mat dstMean, Mat dstStdDev);
 void Mat_Merge(struct Mats mats, Mat dst);
 void Mat_MinMaxLoc(Mat m, double* minVal, double* maxVal, Point* minLoc, Point* maxLoc);
 void Mat_Multiply(Mat src1, Mat src2, Mat dst);
-void Mat_Subtract(Mat src1, Mat src2, Mat dst);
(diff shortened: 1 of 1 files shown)


berak:
yea, but, once other pr's got merged

deadprogram:
This PR still appears to be needed, right?

willdurand:
You could still keep the PRs open and rebase later. It is nice for communicating with others that "this is done" or "working on this".

berak:
hmm but it would mean rebasing *each* outstanding pr if any of those minor things get merged. that's tiring.

i'd rather keep a list of todo's here, and make a pr with half a dozen of them, once the field is clear

willdurand:
There isn't a lot of conflicts to deal with using git so no need to rebase all your PRs, github will tell you when it is needed

berak:
unfortunately, it's a matter of order. the outstanding pr's still have the dupe declaration, so there for sure will be a conflict, if this will get merged first (and why wouldn't anyone merge such a small thing as this first ?)

willdurand:
You could PRs with different patches, and if one PR needs the patch from another one, then you create a single PR for both patches (i.e. commits).

deadprogram:
Generally speaking, git is able to resolve most merges. The problems occur when there are *many* changes to a single file in one PR, and then another comes along to that same file to code that was changed by the more pervasive PR.

The recent core changes had exactly this sort of situation occur, but now I think that should be largely resolved. @berak do you mind if I reopen this PR so I can merge it?

PR #121, opened 2018-03-07
imgcodecs: remove gif extension

berak:
it can neither read or write it (still a license problem), avoid confusion


zak905:
Hi, this is not reflected in the documentation https://godoc.org/gocv.io/x/gocv#FileExt

If I understand well, this should be the reason why I am getting : 

` what():  OpenCV(3.4.2) /tmp/opencv/opencv-3.4.2/modules/imgcodecs/src/loadsave.cpp:910: error: (-2:Unspecified error) could not find encoder for the specified extension in function 'imencode' `

Issue #122, opened 2018-03-07
highgui issues


berak:
1. wrong flag [here](https://github.com/hybridgroup/gocv/blob/master/highgui.go#L39) 2 is only valid for setWindowProp
2. if we use 0 for the flag by default, it will inherit the last (internally saved) size of a window with that name, (e.g. "Hello") which is confusing. imho, the flag should be settable/exposed in the api (and the default flag should be 1, as it is in opencv)
3. the whole class approach. why is this nessecary ? making imshow a classmember is unfortunate, e.g. if you need to visualize a Mat inside a shortlived function, using code like this:

```
func xxx(img gocv.Mat) {
        // do something with img
	window := gocv.NewWindow("Hello2")
	defer window.Close()
	window.IMShow(img)
}
```
we can either skip the Close() and have a memleak, or the window will close immediately after leaving the func. Adding a WaitKey() will block the main thread (and using concurrency for this is not an option, either)

window.WaitKey() might let users think, that they can "select" events for a single window this way, but it isn't so.

willdurand:
> the whole class approach. why is this nessecary ? making imshow a classmember is unfortunate, e.g. if you need to visualize a Mat inside a shortlived function, using code like this:

what would you suggest? I actually like that we can have separate objects.

berak:
> I actually like that we can have separate objects.

it' s somewhat an illusion, isn't it ? (again, waitKey() is global)
does it have any pro's to do so ? (maybe i'm just missing something)

inside c++ opencv, they are indeed objects, they keep a linked list of those, and users "address" a window by it's name

PR #123, opened 2018-03-07
highgui: fix constant in NewWindow

berak:
#122


deadprogram:
So the correct default value should actually be `1`?

deadprogram:
I think you can reopen it now?

berak:
all ok with me, iobviouslywas overly pessimistic

PR #125, opened 2018-03-10
Add Mat.ToImage()

gen2brain:
This PR adds ToImage() to Mat, so it is easy to convert to Go image.Image .

### DIFF
--- a/core.cpp
+++ b/core.cpp
@@ -107,6 +107,11 @@ int Mat_Type(Mat m) {
     return m->type();
 }
 
+// Mat_Step returns the number of bytes each matrix row occupies.
+int Mat_Step(Mat m) {
+    return m->step;
+}
+
 // Mat_GetUChar returns a specific row/col value from this Mat expecting
 // each element to contain a schar aka CV_8U.
 uint8_t Mat_GetUChar(Mat m, int row, int col) {
--- a/core.go
+++ b/core.go
@@ -6,6 +6,7 @@ package gocv
 */
 import "C"
 import (
+	"errors"
 	"image"
 	"image/color"
 	"reflect"
@@ -268,6 +269,11 @@ func (m *Mat) Type() int {
 	return int(C.Mat_Type(m.p))
 }
 
+// Step returns the number of bytes each matrix row occupies.
+func (m *Mat) Step() int {
+	return int(C.Mat_Step(m.p))
+}
+
 // GetUCharAt returns a value from a specific row/col
 // in this Mat expecting it to be of type uchar aka CV_8U.
 func (m *Mat) GetUCharAt(row int, col int) uint8 {
@@ -410,6 +416,56 @@ func (m *Mat) SetDoubleAt(row int, col int, val float64) {
 // in this Mat expecting it to be of type double aka CV_64F.
 func (m *Mat) SetDoubleAt3(x, y, z int, val float64) {
 	C.Mat_SetDouble3(m.p, C.int(x), C.int(y), C.int(z), C.double(val))
+}
+
+// ToImage converts a Mat to a image.Image.
+func (m *Mat) ToImage() (image.Image, error) {
+	t := MatType(m.Type())
+	if t != MatTypeCV8UC1 && t != MatTypeCV8UC3 && t != MatTypeCV8UC4 {
+		return nil, errors.New("ToImage supports only MatType CV8UC1, CV8UC3 and CV8UC4")
+	}
+
+	width := m.Cols()
+	height := m.Rows()
+	step := m.Step()
+	data := m.ToBytes()
+	channels := m.Channels()
+
+	if t == MatTypeCV8UC1 {
+		img := image.NewGray(image.Rect(0, 0, width, height))
+		c := color.Gray{Y: uint8(0)}
+
+		for y := 0; y < height; y++ {
+			for x := 0; x < width; x++ {
+				c.Y = uint8(data[y*step+x])
+				img.SetGray(x, y, c)
+			}
+		}
+
+		return img, nil
+	}
+
+	img := image.NewNRGBA(image.Rect(0, 0, width, height))
+	c := color.NRGBA{
+		R: uint8(0),
+		G: uint8(0),
+		B: uint8(0),
+		A: uint8(255),
+	}
+
+	for y := 0; y < height; y++ {
+		for x := 0; x < step; x = x + channels {
+			c.B = uint8(data[y*step+x])
+			c.G = uint8(data[y*step+x+1])
+			c.R = uint8(data[y*step+x+2])
+			if channels == 4 {
+				c.A = uint8(data[y*step+x+3])
+			}
+			img.SetNRGBA(int(x/channels), y, c)
+		}
+	}
+
+	return img, nil
 }
 
 // AbsDiff calculates the per-element absolute difference between two arrays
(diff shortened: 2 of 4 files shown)


berak:
oh, that looks nice !

* would it be possible, to handle grayscale (1-channel) images, too ? 
(or, at least, have a check for that, and bail out, instead of running into an overflow) 
(it might also happen, that ppl are trying to throw floating point or 16bit images at it)

* a unit (smoke) test for it would make the codecov bot happy

gen2brain:
I added test and now it will return error if type is not correct. It should be possible to also handle grayscale images but needs little investigation.

core.cpp
```diff
@@ -107,6 +107,11 @@ int Mat_Type(Mat m) {
     return m->type();
 }
 
+// Mat_Step returns size of aligned image row in bytes.
```
willdurand: I did not understand this doc, so I googled a bit and found this:

> step returns the number of bytes each matrix row occupies.

Is it the same idea?

core.go
```diff
@@ -268,6 +269,11 @@ func (m *Mat) Type() int {
 	return int(C.Mat_Type(m.p))
 }
 
+// Step returns the size of aligned image row in bytes for this Mat.
```
willdurand: no need for "for this Mat", the receiver is already a Mat.

core.go
```diff
@@ -412,6 +418,49 @@ func (m *Mat) SetDoubleAt3(x, y, z int, val float64) {
 	C.Mat_SetDouble3(m.p, C.int(x), C.int(y), C.int(z), C.double(val))
 }
 
+// ToImage returns Go image.Image from Mat.
```
willdurand: > ToImage converts a Mat to a image.Image.

core.go
```diff
@@ -412,6 +418,49 @@ func (m *Mat) SetDoubleAt3(x, y, z int, val float64) {
...
+	t := m.Type()
+	if t != int(MatTypeCV8U) && t != int(MatTypeCV8UC1) && t != int(MatTypeCV8UC3) && t != int(MatTypeCV8UC4) {
+		return nil, errors.New("ToImage supports only MatType CV8U, CV8UC1, CV8UC3 and CV8UC4")
+	}
+
+	var height, width, channels, step int = m.Rows(), m.Cols(), m.Channels(), m.Step()
+
+	data := m.ToBytes()
+
+	if t == int(MatTypeCV8U) || t == int(MatTypeCV8UC1) {
```
willdurand: Could you tell me what these two types have in common?

I would add a method on Mat instead of doing this comparison here, so that we can reuse it.

core.go
```diff
@@ -412,6 +418,49 @@ func (m *Mat) SetDoubleAt3(x, y, z int, val float64) {
...
+
+		for y := 0; y < height; y++ {
+			for x := 0; x < width; x++ {
+				c.Y = uint8(data[y*step+x])
+				out.SetGray(x, y, c)
+			}
+		}
+
+		return out, nil
+	} else {
```
willdurand: No need for a `else` here, you return in the previous `if`.

core.go
```diff
@@ -412,6 +418,49 @@ func (m *Mat) SetDoubleAt3(x, y, z int, val float64) {
...
+		out := image.NewNRGBA(image.Rect(0, 0, width, height))
```
willdurand: `out` could be named `img`

core.go
```diff
@@ -412,6 +418,49 @@ func (m *Mat) SetDoubleAt3(x, y, z int, val float64) {
...
+		c := color.NRGBA{R: uint8(0), G: uint8(0), B: uint8(0), A: uint8(255)}
```
willdurand: Can you format this on multiple lines?

core.go
```diff
@@ -412,6 +418,49 @@ func (m *Mat) SetDoubleAt3(x, y, z int, val float64) {
...
 }
 
+// ToImage returns Go image.Image from Mat.
+func (m *Mat) ToImage() (image.Image, error) {
+	t := m.Type()
+	if t != int(MatTypeCV8U) && t != int(MatTypeCV8UC1) && t != int(MatTypeCV8UC3) && t != int(MatTypeCV8UC4) {
+		return nil, errors.New("ToImage supports only MatType CV8U, CV8UC1, CV8UC3 and CV8UC4")
+	}
+
+	var height, width, channels, step int = m.Rows(), m.Cols(), m.Channels(), m.Step()
```
willdurand: I am not a huge fan of this. What about:

```
h := m.Rows()
w := m. Cols() 
... 
```

core.go
```diff
@@ -412,6 +418,49 @@ func (m *Mat) SetDoubleAt3(x, y, z int, val float64) {
 	C.Mat_SetDouble3(m.p, C.int(x), C.int(y), C.int(z), C.double(val))
 }
 
+// ToImage returns Go image.Image from Mat.
+func (m *Mat) ToImage() (image.Image, error) {
+	t := m.Type()
+	if t != int(MatTypeCV8U) && t != int(MatTypeCV8UC1) && t != int(MatTypeCV8UC3) && t != int(MatTypeCV8UC4) {
```
willdurand: Maybe we could create a loop instead and:

1. cast the constant only once
2. return the type in the error message (like "mat type is X, expected on of these: {Y, Z}")

willdurand: This also makes me think that the `Type()` function shoukd return `MatType`.

core_test.go
```diff
@@ -685,3 +685,53 @@ func TestMatExtractChannel(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("TestToImage %v.", err)
+	}
+
+	mat5 := NewMatWithSize(101, 102, MatTypeCV8S)
+	defer mat5.Close()
+
+	_, err = mat5.ToImage()
+	if err == nil {
+		t.Error("TestToImage expected error got nil.")
```
willdurand: Can we assert on the error message or rename the variable `matWithUnsupportedType`? It is tricky to understand what's happening here

core.cpp
gen2brain: I changed docs, it was a c/p from include file, just from wrong place.

core.go
gen2brain: Done.

core.go
gen2brain: Actually, I am not sure if MatTypeCV8U is correct, I did a local test with some images I converted with imagemagick, so for now better to remove it.

core.go
gen2brain: Fixed.

core.go
gen2brain: Now `t` is casted to MatType and without MatTypeCV8U it looks better?

core_test.go
gen2brain: Variable renamed.

core.go
gen2brain: Renamed.

gen2brain:
Is it ok now? I didn't squashed commits before.

willdurand:
That looks very very good to me, thank you so much for this 👏

deadprogram:
Hi @gen2brain thank you very much for your contribution. I took the liberty of switching it from the `master` to the `dev` branch as mentioned in our contribution guidelines here: https://github.com/hybridgroup/gocv/blob/master/CONTRIBUTING.md#how-to-use-our-github-repository

Now merging, thanks again!

PR #134, opened 2018-03-12
Mat Type() should return MatType

### LINKED ISSUE #130: Type() should return MatType, not int
I think we should return the specialized type in `Type()`, i.e. `MatType` and not `int`. WDYT ?

willdurand:
Fix #130

---

I believe `Type()` should return a `MatType` because that's what we use
everywhere else, and it would make comparisons more natural. I stumbled upon
this while reviewing #125.

It could be since as a BC break even though we did not reach version `1.0.0` yet.

### DIFF
--- a/core.go
+++ b/core.go
@@ -264,8 +264,8 @@ func (m *Mat) Channels() int {
 }
 
 // Type returns the type for this Mat.
-func (m *Mat) Type() int {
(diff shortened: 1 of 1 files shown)


deadprogram:
Good one @willdurand now merging. Thanks!

PR #135, opened 2018-03-13
core: add more functions

berak:
FindNonZero
Flip
Gemm
Hconcat
Idct
Idft
InsertChannel
Invert
Magnitude
Max


core.go
```diff
@@ -745,6 +745,36 @@ func ExtractChannel(src Mat, dst Mat, coi int) {
 	C.Mat_ExtractChannel(src.p, dst.p, C.int(coi))
 }
 
+// FindNonZero returns the list of locations of non-zero pixels
+//
+// For further details, please see:
+// https://docs.opencv.org/master/d2/de8/group__core__array.html#gaed7df59a3539b4cc0fe5c9c8d7586190
+//
+func FindNonZero(src Mat, idx Mat) {
```
willdurand: It should be `FindNonZero(src, idx Mat)`

core.go
```diff
@@ -745,6 +745,36 @@ func ExtractChannel(src Mat, dst Mat, coi int) {
 	C.Mat_ExtractChannel(src.p, dst.p, C.int(coi))
 }
 
+// FindNonZero returns the list of locations of non-zero pixels
```
willdurand: Missing trailing period.

core.go
```diff
@@ -745,6 +745,36 @@ func ExtractChannel(src Mat, dst Mat, coi int) {
...
+// For further details, please see:
+// https://docs.opencv.org/master/d2/de8/group__core__array.html#gaed7df59a3539b4cc0fe5c9c8d7586190
+//
+func FindNonZero(src Mat, idx Mat) {
+	C.Mat_FindNonZero(src.p, idx.p)
+}
+
+// Flip flips a 2D array around vertical, horizontal, or both axes.
+//
+// FlipCode	a flag to specify how to flip the array; 0 means flipping around the x-axis and positive value (for example, 1) means flipping around y-axis.
```
willdurand: How does that render in godoc?

core.go
```diff
@@ -745,6 +745,36 @@ func ExtractChannel(src Mat, dst Mat, coi int) {
...
+// Negative value (for example, -1) means flipping around both axes.
+//
+// For further details, please see:
+// https://docs.opencv.org/master/d2/de8/group__core__array.html#gaca7be533e3dac7feb70fc60635adf441
+//
+func Flip(src Mat, dst Mat, flipCode int) {
```
willdurand: here too you can skip the first type since the two first arguments have the same type, Go will pick the type after the second argument.

willdurand: I wont repeat the comment but it applies to the whole file :)

core.go
berak: your point..

core.go
berak: what exactly ? a colon behind the param name ?

core.go
berak: understood. my generator isn't yet up to it, and the rest of the codebase isn't too.

willdurand: oh that's smart, so you're using a generator to port all these functions?

core.go
willdurand: the `\t` between the argument name and its description

core.go
berak: lol, i'm trying hard, but it only gets 80% right.

(at least, i can be sure it does not miss params)
(combination of copy/paste stuff done by a machine (they *are* better at it !) and human revision (comments, tests) for now.)

core.go
berak: well that is true for almost any existing function in the codebase.

i'd rather prefer not to make a breaking style change here

berak: start an issue / general discussion about it ?

willdurand: we could but that's how we write functions in Go. We need to fix the other functions if that's the case, but you can start fixing your stuff here :wink:

berak: ok, let's see, how it turns out ..

deadprogram:
Wow! 🔥

PR #136, opened 2018-03-13
Compare Mat Type directly

willdurand:
Since #134, we can directly compare the `Mat` type so no need to cast it anymore (cf. #125).


deadprogram:
Thank you @willdurand

PR #138, opened 2018-03-17
Add HoughCirclesWithParams() func

### LINKED ISSUE #49: Can you give me a sample about the function HoughCircles?
```
gocv.CvtColor(img, edges, gocv.ColorBGRToGray)
gocv.GaussianBlur(edges, edges, image.Pt(9, 9), 2, 2, gocv.BorderDefault)
gocv.HoughCircles(edges, circles, 3, 1.5, float64(edges.Rows()/8))
fmt.Println(circles.Mean(), circles.Rows(), circles.Cols())
```

I want to detect circles.
I can't use the HoughCircles because no samples for HoughCircles.

How to use the `circles`?

please help me, tks

### LINKED ISSUE #52: add param1, param2, minRadius, maxRadius params to HoughCircles
Currently `HoughCircles` takes in 4 parameters
```func HoughCircles(src Mat, circles Mat, method int, dp float64, minDist float64)```

the OpenCV HoughCircles methods lets you pass in ```double param1, double param2, int minRadius, int maxRadius``` as well as the 4 params currently present.

any chance HoughCircles will be updated to allow for the extra 4 params?

### PR DESCRIPTION
**Pull request by willdurand**
Fix #49
Fix #52
Refs #105

---

This PR adds the `HoughCirclesWithParams()` func, adding all the arguments
of the opencv function. ~According to @berak, it is useless to have such an
implementation without min/max radius so I chose to fix the existing `func`
rather than adding a `WithParams` version.~

It also provides an example for those who want to know how to use this function.

screen shot 2018-03-17 at 10 31 56

### CODE REVIEW

File: cmd/find-circles/main.go [hunk 1]
```diff
@@ -0,0 +1,78 @@
...
+	)
+
+	blue := color.RGBA{0, 0, 255, 0}
+	red := color.RGBA{255, 0, 0, 0}
+
+	ch := circles.Channels()
+	for i := 0; i < circles.Cols(); i++ {
+		x := int(circles.GetFloatAt(0, i*ch))
+		y := int(circles.GetFloatAt(0, i*ch+1))
+		r := int(circles.GetFloatAt(0, i*ch+2))
```
willdurand: Maybe we could create a data structure for this, like `Vec3f` in opencv, and we could have a new `GetVecFAt()` func to return a vector of floats.

```
// VecF is a generic vector of floats.
type VecF []float32
```

WDYT?

willdurand: Here is the diff:

```diff
diff --git a/cmd/find-circles/main.go b/cmd/find-circles/main.go
index a39c1af..7c7d2bc 100644
--- a/cmd/find-circles/main.go
+++ b/cmd/find-circles/main.go
@@ -58,11 +58,11 @@ func main() {
 	blue := color.RGBA{0, 0, 255, 0}
 	red := color.RGBA{255, 0, 0, 0}
 
-	ch := circles.Channels()
 	for i := 0; i < circles.Cols(); i++ {
-		x := int(circles.GetFloatAt(0, i*ch))
-		y := int(circles.GetFloatAt(0, i*ch+1))
-		r := int(circles.GetFloatAt(0, i*ch+2))
+		v := circles.GetVecfAt(0, i)
+		x := int(v[0])
+		y := int(v[1])
+		r := int(v[2])
 
 		gocv.Circle(cimg, image.Pt(x, y), r, blue, 2)
 		gocv.Circle(cimg, image.Pt(x, y), 2, red, 3)
diff --git a/core.go b/core.go
index 0806e45..073b769 100644
--- a/core.go
+++ b/core.go
@@ -1070,6 +1070,22 @@ type KeyPoint struct {
 	Octave, ClassID       int
 }
 
+// Vecf is a generic vector of floats.
+type Vecf []float32
+
+// GetVecfAt returns a vector of floats. Its size corresponds to the number of
+// channels of the Mat.
+func (m *Mat) GetVecfAt(row int, col int) Vecf {
+	ch := m.Channels()
+	v := make(Vecf, ch)
+
+	for c := 0; c < ch; c++ {
+		v[c] = m.GetFloatAt(row, col*ch+c)
+	}
+
+	return v
+}
+
 func toByteArray(b []byte) C.struct_ByteArray {
 	return C.struct_ByteArray{
 		data:   (*C.char)(unsafe.Pointer(&b[0])),
```

deadprogram: I had initially looked at an implementation like you suggest @willdurand for `Vec3f` however did not keep going with it. There are quite a few different vector types used within OpenCV for different filters and algorithms, so at the time I elected to use the `GetXXXAt()` and `SetXXXAt()` functions to avoid what at the time seemed like would require a lot of code duplication. We would need 2, 3 and 4-d versions for all of the OpenCV unsigned/signed ints and various sized floats.

That said, it is still a finite set, and there are many advantages to reducing the code needed for the consumer of the package. Anyone else want to comment here? I'd like to hear more viewpoints here, if possible.

willdurand: I could open an issue for that then, or a separate PR to avoid blocking this one.

As you can see in the diff above, the `GetVecfAt()` returns a vector of floats whose size depends on the number of channels, therefore it should be flexible enough. Of course, we would need `GetVecdAt()` for `double/flaot64`, `GetVeciAt()` for `int`, etc.

deadprogram: One disadvantage that occurs to me, is that it requires copying all of the data from the entire `Mat` into the `Vecf` vs. the current implementation that just accesses the original data.

willdurand: right, although it provides a simple function to access the data, which can be used for less critical apps. It took me a while to figure out how to retrieve the data for the example code (because all SO/Internet answers talk about a vector or `Vec3f`).

deadprogram: I agree that a separate issue/PR would be best for this vector conversation.

deadprogram:
So, would this function not be better implemented as a `HoughCirclesWithParams()`? The distinction is simply if OpenCV has any default params for a function, we have been implementing 2 versions of the function since Go does not support overloading. So for an OpenCV function: 

```c
opencv::xYZ(int p1, int p2, int p3=2, int p4=3);
```

We would define 2 functions in GoCV:

```go
// uses default param values
XYZ(p1, p2)

// sets each param
XYZWithParams(p2, p2, p3, p4)
```

Hopefully that clarifies the intent behind the design decision.

willdurand:
@deadprogram that's what I wanted to do at first but @berak said somewhere that there was little value in having `HoughCircles()` without all the arguments. I'd be happy to create a new function `WithParams` if the existing one does something useful.

deadprogram:
It does whatever it does with the default params, as any of the other OpenCV functions with default params. I would say the default params are useful, but not sufficient. In other words, both are needed.

willdurand:
> It does whatever it does with the default params, as any of the other OpenCV functions with default params. I would say the default params are useful, but not sufficient. In other words, both are needed.

Let's add this `WithParams` function then 🤘

willdurand:
Ok PR updated with the new function.

deadprogram:
Thanks @willdurand now merging.

Issue #137, opened 2018-03-13
Use Go-style return values for gocv functions that have result Mat


willdurand:
When reviewing #135, I was like "wait, why is this test actually passing?". The test was about flipping a `Mat` into a brand new `Mat`, both passed to a function that looks roughly like this:

```go
func Flip(src Mat, dst Mat, flipCode int) {}
```

In Go, parameters passed by value (default like above) should not be modified (and in any case, it will never change the value in the parent block). In gocv, this would works though because we manipulate a C pointer under the hood but this is a bit misleading.

Question is: could we agree on a "convention" stating that all arguments modified should be passed by pointer?

If we take the `Flip()` function as an example:

```go
src := ...
dst := gocv.NewMat()

// wait, how does `dst` can be updated? No return-value either? Hmmm.
gocv.Flip(src, dst, 0);

// versus

// oh right so `Flip` takes a `src` Mat and puts the result in `dst`. Lovely <3
gocv.Flip(src, &dst, 0);
```

berak:
somehow, you've hit a spot on the general concept here.

the gocv wrappers all go like this:

    type someclass {
         p pointer_to_some_CGO_implementation
    }
  
so, actually, whatever p points to, will get changed, not the go instance.
(and taking the address of it, won't mend it)

imho, it's already a problem in the underlying opencv c++ code, [for further midbending fun, have a look here]( https://github.com/opencv/opencv/blob/master/modules/core/include/opencv2/core/mat.hpp#L152-L421)

willdurand:
> somehow, returning dst, instead of taking it by a (invisible) reference, like it is done in python, would be a better idea. (too late, too change it for all of the core bindings ?)

I think returning values would be indeed better.

berak:
right, yet it still *works*, like it is, now

deadprogram:
There was a discussion about this at some point on slack or something like that. The question is, should we be more like the C++ API (the current way) or more like the Python API (return a new Mat in function that has a 'dest' that the dev needs to close).

I am still not pulled either way with great conviction, the C++ style seemed the best at the time, but we could decide to change the interface. Seems like doing so sooner would be best if we are going to do so.

deadprogram:
As I recall it was @221bytes that I was discussing this with on the Golang Slack channel.

In any case, we can decide to follow one of 2 approaches here:

[...]

PR #139, opened 2018-03-20
Add GetVecfAt() and GetVecifAt() functions to Mat

willdurand:
This is a proposal to simplify the way we get data from `Mat`, which we use as
a general purpose data structure. I find it more convenient because when
looking on the Internet (for example, for Hough transforms), people talk about
`Vec` or `Vec3f` (and so on), not `Mat`. Of course, a matrix can be used as a
vector, but it is less simple than having an accessor for that.

It is still possible to use the `Mat` and do the maths by yourself, it is only
a nice helper. We could add `GetVecdAt()` for `double`/`float64`, `GetVeciAt()`
for `int`, etc.

### DIFF
--- a/cmd/README.md
+++ b/cmd/README.md
@@ -26,6 +26,10 @@ Captures video from a connected camera, then uses the CascadeClassifier to detec
 
 Captures video from a connected camera, then uses the CascadeClassifier to detect faces, and draw a rectangle around each of them, before displaying them within a Window
 
+## Find-circles
+
+Find circles in an image using the Hough transform.
+
 ## MJPEG-Streamer
 
 Opens a video capture device, then streams MJPEG from it that you can view in any browser.
--- a/cmd/find-circles/main.go
+++ b/cmd/find-circles/main.go
@@ -58,11 +58,11 @@ func main() {
 	blue := color.RGBA{0, 0, 255, 0}
 	red := color.RGBA{255, 0, 0, 0}
 
-	ch := circles.Channels()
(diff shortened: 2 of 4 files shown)


deadprogram:
Hi @willdurand was thinking about this PR in the context of the comments made by @maruel on #137 and two thoughts spring to mind.

- Might be great to possibly incorporate some of his ideas on typed `Mat` here
- I'd really like to see an implementation that did not require copying the data. Perhaps returning a function instead of a slice? For example:

```go
// GetVecfAt () returns func(idx int)(val float32){...}
v := circles.GetVecfAt(0, i)
x := int(v(0))
```

maruel:
I think copying 16 bytes is way faster than 1) creating a function closure and 2) calling a function; calling a function in Go is somewhat expensive, and creating a function closure literally creates assembly on the heap.

A micro benchmark could be helpful if you care about perf since the costs are hard to guess and guts feelings may be misleading.

deadprogram:
Good point, @maruel

willdurand:
So what do you want me to do?

deadprogram:
@willdurand I suppose that means carry on with your plans to implement the wrappers around `VecX` for the various data types. :horse_racing:

willdurand:
Should we add them all here or only when needed?

deadprogram:
Maybe do `Veci` and then the rest as needed? That seems to the other most common type.

willdurand:
@deadprogram done :)

deadprogram:
Thanks @willdurand you are on a roll! Merging...

PR #143, opened 2018-03-22
Return Go Error for Read()/Write() operations

deadprogram:
As suggested by @maruel in #137 always return Go `Error` for all Read or Write operations instead of just `Bool`. This is the second breaking API change on the road to the 1.x API.

Here is an example of it in use:

```go
if err := webcam.Read(&img); err != nil {
    fmt.Printf("cannot read device %d\n", deviceID)
    return
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ import (
 
 func main() {
 	webcam, _ := gocv.VideoCaptureDevice(0)
-	window := gocv.NewWindow("Hello")	
+	window := gocv.NewWindow("Hello")
 	img := gocv.NewMat()
 
 	for {
@@ -87,7 +87,7 @@ func main() {
 
 	fmt.Printf("start reading camera device: %v\n", deviceID)
 	for {
-		if ok := webcam.Read(&img); !ok {
+		if err := webcam.Read(&img); err != nil {
 			fmt.Printf("cannot read device %d\n", deviceID)
 			return
 		}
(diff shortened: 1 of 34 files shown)


README.md
```diff
@@ -87,7 +87,7 @@ func main() {
 
 	fmt.Printf("start reading camera device: %v\n", deviceID)
 	for {
-		if ok := webcam.Read(&img); !ok {
+		if err := webcam.Read(&img); err != nil {
 			fmt.Printf("cannot read device %d\n", deviceID)
```
maruel: You'd want to print out the error message.

videoio.go
```diff
@@ -204,8 +205,11 @@ func (v *VideoCapture) IsOpened() bool {
 
 // Read reads the next frame from the VideoCapture to the Mat passed in
 // as the param. It returns false if the VideoCapture cannot read frame.
-func (v *VideoCapture) Read(m *Mat) bool {
-	return C.VideoCapture_Read(v.p, m.p) != 0
+func (v *VideoCapture) Read(m *Mat) (err error) {
+	if C.VideoCapture_Read(v.p, m.p) == 0 {
+		err = errors.New("VideoCapture Read error")
```
maruel: Oh, the library doesn't return an enum for error description to be able to describe the error at least a bit? :/

deadprogram: OpenCV returns a bool here, so not much we can do about it.

maruel: Oh I see, this is kinda disappointing. I wonder if there is much value into the PR then for the code churn it represents. I naively thought there would be some form of error code. :/

deadprogram: Agreed. I will probably just close it then to avoid the hassle to everyone.

PR #147, opened 2018-03-23
Update links in caffe-classifier demo

willdurand:
I wanted to try this demo and the links were outdated.

Maybe we could host these files in a separate repo?


deadprogram:
You verified that the sample runs correctly with these files?

willdurand:
> You verified that the sample runs correctly with these files?

yup, it does. `prototxt` is the same file and the other one has a different name but I found it in the updated opencv doc.

deadprogram:
OK, thanks for the update.

willdurand:
you're welcome :)

PR #151, opened 2018-03-24
Make borderType in CopyMakeBorder be type BorderType

milosgajdos83:
CopyMakeBorder currently accepts int in place of borderType. It would be nice if `CopyMakeBorder` accepted `BorderType` instead of `int`. This way the parameter would be more consistent with `OpenCV` and other `gocv` function that accept similar parameters and it would also be user friendlier to developers i.e. more straightforward to understand.


willdurand:
it is hard to see what has changed, could you please rebase your branch? It seems you have unwanted modifications in your commit.

milosgajdos83:
Uh, you're right! I think I know what happened! I haven't synced my fork with gocv 🙄
Let me rebase

deadprogram:
Thanks @milosgajdos83 now merging.

PR #156, opened 2018-04-01
Implementing Mat.FromImage helper function

denismakogon:
- turns golang image.Image to gocv.Mat
 - "versa" function to the Mat.ToImage

PR #158, opened 2018-04-02
Implementing Mat.FromImage helper function

denismakogon:
Was: #156

### DIFF
--- a/core.go
+++ b/core.go
@@ -499,35 +499,33 @@ func (m *Mat) ToImage() (image.Image, error) {
 }
 
 // FromImage converts go Image to the Mat object
-func (m *Mat) FromImage(img image.Image, imgType ImageMagicType, mt MatType) error {
+func FromImage(img image.Image, imgType ImageMagicType, mt MatType) (*Mat, error) {
 	bounds := img.Bounds()
 	buf := new(bytes.Buffer)
 	switch imgType {
 	case JPEG:
 		err := jpeg.Encode(buf, img, nil)
 		if err != nil {
-			return err
+			return nil, err
 		}
 	case PNG:
 		err := png.Encode(buf, img)
 		if err != nil {
-			return err
+			return nil, err
 		}
 	case GIF:
 		err := gif.Encode(buf, img, nil)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


core.go
```diff
@@ -486,35 +486,33 @@ func (m *Mat) ToImage() (image.Image, error) {
 }
 
 // FromImage converts go Image to the Mat object
-func (m *Mat) FromImage(img image.Image, imgType ImageMagicType, mt MatType) error {
+func FromImage(img image.Image, imgType ImageMagicType, mt MatType) (*Mat, error) {
```
deadprogram: Seems like should probably just return a `Mat` not a `*Mat`.

deadprogram:
Hi @denismakogon looks like the tests are failing now on both Linux and Windows, just in case you had not noticed.

denismakogon:
Hi! @deadprogram, I've seen that didn't have time to investigate those, so, probably would need certain help with that if possible.

deadprogram:
Hi @denismakogon now on top of the failing tests there is also a merge conflict for this PR. Seems like it will need some more love in order to land...

deadprogram:
Seems like the approach in #228 is more efficient @denismakogon what do you think?

denismakogon:
looks good, will try updating PR ASAP.

edwvee:
Have looked code through. Tried common thing:
```go
//img is finely decoded jpeg
bounds := img.Bounds()
	x := bounds.Dx()
	y := bounds.Dy()
	data := []byte{}
	w := bytes.NewBuffer(data)
	jpeg.Encode(w, img, &jpeg.Options{Quality: 90})
	data = w.Bytes()
	mat, err := gocv.NewMatFromBytes(y, x, gocv.MatTypeCV8UC3, data)

	if err != nil {
		log.Fatal(err)
	}
```
No error. But when i launched ```mat.ToImage()```  - received panic. Who or what is wrong: the way I uderstood this PR's code, this PR or mat.ToImage()?

deadprogram:
This PR has now been superseded by #234 which has been merged into the `dev` branch. Now closing.

PR #164, opened 2018-04-10
Fix: It doesn't compile on my machine.

DesWurstes:


willdurand:
mmh would you mind adding some context? I dont see how this could work? I am not a cpp expert but structure initialization was done correctly before your patch if I am not mistaken.

DesWurstes:
This is valid in C++ but is invalid in C:
```
someStruct{5,6};
```
This is valid in both (also, is a standard. I'm not sure if the first example was ISO):
```
(someStruct){5,6};
```

More information about it:

[Compound Literals](https://gcc.gnu.org/onlinedocs/gcc/Compound-Literals.html) manual:

```c
static struct foo x = (struct foo) {1, 'a', 'b'};
static int y[] = (int []) {1, 2, 3};
static int z[] = (int [3]) {1};
```
> The above lines are equivalent to the following: 
```c
static struct foo x = {1, 'a', 'b'};
static int y[] = {1, 2, 3};
static int z[] = {1, 0, 0};
```
Also see this: https://stackoverflow.com/a/330828/

and https://stackoverflow.com/a/330824/6557621

deadprogram:
Hi @DesWurstes just wondering what OS and compiler you are using? 

Also, did you run `env.sh` to setup the expected CGO env vars, or do you have some other way to set them? I was under the impression that https://github.com/hybridgroup/gocv/blob/master/env.sh#L26 sets the needed compiler version.

deadprogram:
Looking more closely, I see that this PR eliminates some compiler warnings, and very likely @DesWurstes you are using strict compiler settings that fail on these warnings.

That certainly seems like a good thing! 

I changed the branch for this PR to the `dev` branch as discussed here https://github.com/hybridgroup/gocv/blob/master/CONTRIBUTING.md#how-to-use-our-github-repository and I will now merge.

Thanks @DesWurstes

PR #166, opened 2018-04-11
Feature/core

kylycht:
PR adds:
- Vconcat(Mat src1, Mat src2, Mat dst)
- Rotate(Mat src, Mat dst, int rotateCode)

### DIFF
--- a/core.cpp
+++ b/core.cpp
@@ -344,6 +344,14 @@ void Mat_Hconcat(Mat src1, Mat src2, Mat dst) {
     cv::hconcat(*src1, *src2, *dst);
 }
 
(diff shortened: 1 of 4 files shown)


kylycht:
@deadprogram missed it, test added

deadprogram:
Thank you @kylycht test looks good. I just noticed that both of these functions are actually in the 'core` module. Perhaps you would not mind moving `Rotate` into `core.go` etc.? Sorry I did not realize that before on my last feedback.

Thanks again for contributing!

kylycht:
@deadprogram sure, added everything to core module(squashed commits as well)

thanks for the great repo!

deadprogram:
Thanks @kylycht now merging.

PR #167, opened 2018-04-12
Add contrib/img_hash module

special:
I went on a bit of a tangent this evening.

This wraps almost all of the img_hash module. The API is largely equivalent: `ImgHashBase` is an interface with Compute and Compare methods, and each hash has a struct implementing that interface.

Differences are:

* The non-class convenience functions for compute are not implemented, because their names would collide and the lack of an equivalent compare made them sort of pointless.
* Hashes that take parameters have these as public members. To solve the default value problem, those hashes (but inconsistently, only those hashes) have a `NewSomethingHash` function.
* Several functions for BlockHash and RadialVarianceHash aren't implemented because they would require keeping state, which would require adding a `Close()` function or a great deal of unnecessary copying. A reasonable solution might be to give those custom compute functions that return the extra data. I have no idea how important this is.
* img_hash is somewhat documented, which is more than I can say.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ before_cache:
 
 script:
   - export CGO_CPPFLAGS="-I${HOME}/usr/include"
-  - export CGO_LDFLAGS="-L${HOME}/usr/lib -lopencv_core -lopencv_videoio -lopencv_face -lopencv_imgproc -lopencv_highgui -lopencv_imgcodecs -lopencv_objdetect -lopencv_features2d -lopencv_video -lopencv_dnn -lopencv_xfeatures2d -lopencv_plot -lopencv_tracking"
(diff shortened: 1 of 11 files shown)


deadprogram:
Hi @special this looks very cool. Can you please add the `img_hash` module to the Travis build here:

https://github.com/hybridgroup/gocv/blob/master/.travis.yml#L44

and the AppVeyor build here:

https://github.com/hybridgroup/gocv/blob/master/appveyor.yml#L29

Also the `env.sh`:
https://github.com/hybridgroup/gocv/blob/master/env.sh

and `env.cmd`:
https://github.com/hybridgroup/gocv/blob/master/env.cmd

Thank you for the contribution!

special:
Fixed all of the build environment issues (well, we'll see what CI says).

I also added a second commit with an `img-similarity` example:

```
$ ./img-similarity -all ../../images/space_shuttle.jpg ../../images/toy.jpg
PHash: similarity 32
AverageHash: similarity 32
BlockMeanHash: similarity 117
BlockMeanHash: similarity 443
ColorMomentHash: similarity 42.20263981441593
MarrHildrethHash: similarity 256
RadialVarianceHash: similarity 0.49646512207080923
```

The values are more or less meaningless unless you research each algorithm, but it's a reasonable demo and maybe a useful tool for comparing their results.

deadprogram:
Thank you @special now merging.

Issue #159, opened 2018-04-03
[Question] How to use Scalar and InRange


milosgajdos83:
I'm working on a simple program where I'm trying to cut out pizza laid over blue background. 

<img width="325" alt="pizza" src="https://user-images.githubusercontent.com/1392526/38222677-20cc1a7c-36de-11e8-96c5-a2026848830b.png">

I figured it could hack something up using `InRange` and then apply the mask over the original image:

```go
       // read image
       pizzaPath := filepath.Join("pizza.png")
        pizza := gocv.IMRead(pizzaPath, gocv.IMReadColor)
        if pizza.Empty() {
                fmt.Printf("Failed to read image: %s\n", pizzaPath)
                os.Exit(1)
        }
        // Convert BGR to HSV image (dont modify the original)
        hsvPizza := gocv.NewMat()
        gocv.CvtColor(pizza, &hsvPizza, gocv.ColorBGRToHSV)
        lower := gocv.NewMatFromScalar(gocv.NewScalar(110.0, 100.0, 100.0, 0.0), gocv.MatTypeCV8U)
        upper := gocv.NewMatFromScalar(gocv.NewScalar(130.0, 255.0, 255.0, 0.0), gocv.MatTypeCV8U)
        // Blue mask
        mask := gocv.NewMat()
        gocv.InRange(hsvPizza, lower, upper, &mask)
        // cut out pizza mask
        pizzaMask := gocv.NewMat()
        gocv.Merge([]gocv.Mat{mask, mask, mask}, &pizzaMask)
        // cut out the pizza and convert back to BGR
        gocv.BitwiseAnd(hsvPizza, pizzaMask, &hsvPizza)
        gocv.CvtColor(hsvPizza, &hsvPizza, gocv.ColorHSVToBGR)
        // write image to filesystem
        outPizza := "no_pizza.jpeg"
        if ok := gocv.IMWrite(outPizza, hsvPizza); !ok {
                fmt.Printf("Failed to write image: %s\n", outPizza)
                os.Exit(1)
        }
```

However it seems that I must be missing out something with regards to how to use `gocv.Scalar` type or `gocv.InRange()` function, but I'm not quite sure what as the resulting image is almost all black with a slight hint of an edge of the pizza in the middle of the image. What I would expect is a blue image with a black hole in the middle of it i.e. the pizza should be masked out as it's not blue.

milosgajdos83:
[...]

PR #178, opened 2018-04-29
Adding InScalarRange and NewMatFromScalarWithSize functions

milosgajdos83:
This commit adds two functions which help when dealing with `Scalar`s:
* `InScalarRange(src Mat, lb, ub Scalar, dst *Mat)`
* `NewMatFromScalarWithSize(s Scalar, rows int, cols int, mt MatType) Mat`

OpenCV C++ API allows to pass in `Scalar` to `cv::inRange` thanks to C++ templating capabilities, however due to the lack of generics and/or function overloading in `Go` I had to add a separate function (`InScalarRange`) to make the same functionality available in core `gocv` API

In order to make some useful operations on images using simpler code (as reported in #159) `NewMatFromScalarWithSize` allows to create a new `Mat` from a specified `Scalar` with specified dimensions.


deadprogram:
Hi @milosgajdos83 thanks for the PR. Looks good, my only real comment is that I think the function should probably be named `InRangeWithScalar()` to help those performing a search in GoCV GoDocs for "inrange".

What do you think?

milosgajdos83:
Excellent point @deadprogram - makes perfect sense.

I've now renamed the function as per your recommendation.

deadprogram:
Thanks, that looks good @milosgajdos83 

Just one more thing... I wonder if perhaps you should change `NewMatFromScalarWithSize' to be `NewMatWithSizeFromScalar" anf then match the params order to be same as https://docs.opencv.org/3.4.1/d3/d63/classcv_1_1Mat.html#a3620c370690b5ca4d40c767be6fb4ceb

milosgajdos83:
Great attention to detail @deadprogram 

Changed the names -- really should have thought of that myself in the previous commit.

Cheers.

deadprogram:
Thanks @milosgajdos83 for indulging me. Now merging.

Issue #162, opened 2018-04-07
Make fail on ubuntu 17.10


muka:
Hi
`make install` fails on ubuntu 17.10 and fallback to fedora(?) `dnf` install. I do know if would make sense to handle this also for ubuntu 18 which is going to  be released

Also `libjasper-dev` is unavaliable in ubuntu after 17.04
A workaround is here https://stackoverflow.com/a/48326450/833499

```
sudo add-apt-repository "deb http://security.ubuntu.com/ubuntu xenial-security main"
sudo apt update
sudo apt install libjasper-dev
```

Thanks!

Auenc:
Also occurs on Ubuntu 16.04

deadprogram:
Any one able to create a PR with changes to the Makefile to accommodate this?

maruel:
I think disabling jpeg2000 support in opencv is a saner solution (nobody uses this format) than modifying the host's apt configuration to fetch from old releases (and not a backport proper).

Testing locally with `sudo apt remove libjasper-dev` then doing a full rebuild with `WITH_JASPER=OFF`

muka:
Yes I agree, your solution is saner definitely!

sent from my phone

Il mar 1 mag 2018, 14:36 M-A <notifications@github.com> ha scritto:

hubyhuby:
+1
I can confirm the same issue on my Ubuntu 17.10
make deps
make deps_fedora
sudo: dnf: command not found
Makefile:32: recipe for target 'deps_fedora' failed
make[1]: *** [deps_fedora] Error 1

...

deadprogram:
Using the `dev` branch, @hubyhuby or the `master`?

gibeautc:
With 18.04 the Makefile seems to think the system is Fedora..... allthought it seems the commands the Makefile uses are still returing correclty.  I was able to modifiy it,and removed the install of of jasper-dev, now stuck on 
/sbin/ldconfig.real: /usr/local/lib/libhidapi-libusb.so.0 is not a symbolic link

/sbin/ldconfig.real: /usr/local/lib/libhidapi-hidraw.so.0 is not a symbolic link

/sbin/ldconfig.real: /usr/lib/libhidapi-libusb.so.0 is not a symbolic link

/sbin/ldconfig.real: /usr/lib/libhidapi-hidraw.so.0 is not a symbolic link

comes up for both make deps, and make build

PR #179, opened 2018-04-30
fix ubuntu install error and missing deps note

muka:
This PR address #162 

- on Ubuntu ifneq doesn't want empty apostrophe, to be tested with other platform also (https://www.gnu.org/software/make/manual/html_node/Conditional-Syntax.html)
- Added a note in readme regarding missing `libjasper`. Not sure if this apply to 18.04 also, can add it to Makefile in case

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,10 +15,10 @@ test:
 	go test .
 
 deps:
-ifneq ($(IS_FEDORA),'')
+ifneq ($(IS_FEDORA),)
 	$(MAKE) deps_fedora
 else
-ifneq ($(IS_RH_CENTOS),'')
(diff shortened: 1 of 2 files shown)


deadprogram:
Hi @muka thanks for this. 

Is there anyone who can test the Makefile with these changes on a distro other than Ubuntu?

Also, I think adding a `make jasper` task to the Makefile would be a good approach.

Lastly, please switch the branch for your PR over to `dev` as mentioned here: https://github.com/hybridgroup/gocv/blob/master/CONTRIBUTING.md#how-to-use-our-github-repository

Thank you!

muka:
I have rewritten the PR, closing this for #181

PR #180, opened 2018-04-30
Improve Makefile

maruel:
- The distro detection didn't work on Ubuntu, it defaulted to Fedora.
- Change it to not be recursive and just be a normal make dependency instead.
- Do not lock make paralelism to 4, some systems may have a LOT of CPUs. :)
- Make the opencv version a variable, so it can be updated easily and overridden
  at the command line.
- Make the temporary directory configurable.
- Move the sudo commands into sudo_install.
- Run make preinstall explicitly first so sudo make install doesn't create root
  owned file in the temporary directory, which made the directory undeletable as
  a user.
- Change test to run tests for ./contrib too

Additional notes:
- Installing pvl is not done here, as this change is already large enough.
- TestVideoCaptureFile fails on systems without a web cam; it still exists :)
  That would require a concept of smoke tests but that's out of scope here.
- The instructions could now read:
    make install
- The defaults can now easily be updated. For example, one can download, build
  and install OpenCV 3.4 (instead of 3.4.1) using the current directory as the
  temporary directory with:
    OPENCV_VERSION=3.4 TMP_DIR= make install

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,29 +1,35 @@
 .ONESHELL:
 .PHONY: test deps download build clean astyle cmds
 
+# OpenCV version to use.
+OPENCV_VERSION?=3.4.1
+
+# Temporary directory to put files into.
+TMP_DIR?=/tmp/
 
 # Package list for each well-known Linux distribution
 RPMS=make cmake git gtk2-devel pkg-config libpng-devel libjpeg-devel libtiff-devel tbb tbb-devel libdc1394-devel jasper-libs jasper-devel
 DEBS=unzip build-essential cmake git libgtk2.0-dev pkg-config libavcodec-dev libavformat-dev libswscale-dev libtbb2 libtbb-dev libjpeg-dev libpng-dev libtiff-dev libjasper-dev libdc1394-22-dev
 
 # Detect Linux distribution
-IS_FEDORA=$(shell which dnf 2>/dev/null)
-IS_DEB_UBUNTU=$(shell which apt-get 2>/dev/null)
-IS_RH_CENTOS=$(shell which yum 2>/dev/null)
+distro_deps=
+ifneq ($(shell which dnf 2>/dev/null),)
+	distro_deps=deps_fedora
+else
+ifneq ($(shell which apt-get 2>/dev/null),)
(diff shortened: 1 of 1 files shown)


maruel:
Sadly on my ODROID-C1+ running Ubuntu 16.06, I get:
```
~/go/src/gocv.io/x/gocv$ go test
# gocv.io/x/gocv
./imgproc.go:39: type [1073741824]_Ctype_struct_Point larger than address space
./imgproc.go:39: type [1073741824]_Ctype_struct_Point too large
./imgproc.go:769: type [1073741824]_Ctype_struct_Point larger than address space
./imgproc.go:769: type [1073741824]_Ctype_struct_Point too large
./imgproc.go:1040: type [1073741824]_Ctype_struct_Point larger than address space
./imgproc.go:1040: type [1073741824]_Ctype_struct_Point too large
FAIL	gocv.io/x/gocv [build failed]
```

deadprogram:
Hi @maruel this is seriously great. A few items:

- Regarding the error on your ODROID-C1+ are you perhaps running a 32-bit version of the OS? This error appears to be because of allocating an array to hold the returned CGo pointer that exceeds the available size https://github.com/hybridgroup/gocv/blob/master/imgproc.go#L39

How should this be handled, do you think? Conditional code for 32/64 bit systems? Allocate something smaller?

- We probably want to add a make task for adding the repo for jasper similar to what @muka just proposed in https://github.com/hybridgroup/gocv/pull/181/files#diff-b67911656ef5d18c4ae36cb6741b7965R38

maruel:
For the 32 bits issue, file https://github.com/hybridgroup/gocv/issues/182 so I can address it separately.

Improved the makefile to address all the current concerns, and a bit more. :)

jpeg2000 was really cool in 1999 (I used it back then) but it never caught on.

deadprogram:
@maruel I did not realize we were dragging around jasper just for such an ancient format. No wonder Ubuntu dropped from distro.

We need to add that same flag to the docs in the README and on the website, so others can realize what is and is not going on with the current build process.

Now merging, thanks again for this terrific contribution.

PR #184, opened 2018-05-02
env.cmd: convert to CRLF

maruel:
Since it is a Windows specific file, it is preferable to keep it as CRLF. This
makes it editable in notepad.


deadprogram:
Thanks @maruel your attention to detail is impressive.

Issue #108, opened 2018-02-28
Failed assertion (!empty()) in detectMultiScale when running example


andybons:
Environment:

```sh
gocv version: 0.9.0
opencv lib version: 3.4.0

$ gcc --version
Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.13.sdk/usr/include/c++/4.2.1
Apple LLVM version 9.0.0 (clang-900.0.39.2)
Target: x86_64-apple-darwin17.4.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

$ sw_vers
ProductName:    Mac OS X
ProductVersion: 10.13.3
BuildVersion:   17D102
```

When running the example from https://gocv.io/writing-code/face-detect/ (the https://gocv.io/writing-code/hello-video/ ran perfectly) I see my camera’s green light go on, then the program crashes:

```
start reading camera device: 0
OpenCV Error: Assertion failed (!empty()) in detectMultiScale, file /tmp/opencv-20180113-80339-d8g5mw/opencv-3.4.0/modules/objdetect/src/cascadedetect.cpp, line 1698
libc++abi.dylib: terminating with uncaught exception of type cv::Exception: /tmp/opencv-20180113-80339-d8g5mw/opencv-3.4.0/modules/objdetect/src/cascadedetect.cpp:1698: error: (-215) !empty() in function detectMultiScale

SIGABRT: abort
PC=0x7fff599bbe3e m=0 sigcode=0
signal arrived during cgo execution

goroutine 1 [syscall, locked to thread]:
runtime.cgocall(0x409a530, 0xc420059dd8, 0xc420000180)
        /usr/local/go/src/runtime/cgocall.go:128 +0x64 fp=0xc420059da8 sp=0xc420059d70 pc=0x40047f4
gocv.io/x/gocv._Cfunc_CascadeClassifier_DetectMultiScale(0xa404c90, 0xa4500a0, 0x0, 0x0)
        _cgo_gotypes.go:674 +0x4d fp=0xc420059dd8 sp=0xc420059da8 pc=0x409604d
[...]

Issue #68, opened 2018-01-28
MacOS High Sierra 10.13.2 error while face detection 


0just0:
I have installed the library and run a couple of examples successfully but got this error message while running face detection code from documentation

```start reading camera device: 0
OpenCV Error: Assertion failed (!empty()) in detectMultiScale, file /tmp/opencv-20180113-80339-d8g5mw/opencv-3.4.0/modules/objdetect/src/cascadedetect.cpp, line 1698
libc++abi.dylib: terminating with uncaught exception of type cv::Exception: /tmp/opencv-20180113-80339-d8g5mw/opencv-3.4.0/modules/objdetect/src/cascadedetect.cpp:1698: error: (-215) !empty() in function detectMultiScale

SIGABRT: abort
PC=0x7fff6dbc6e3e m=0 sigcode=0
signal arrived during cgo execution

goroutine 1 [syscall, locked to thread]:
runtime.cgocall(0x409ef30, 0xc420057dd0, 0xc420057df8)
	/usr/local/Cellar/go/1.9.2/libexec/src/runtime/cgocall.go:132 +0xe4 fp=0xc420057da0 sp=0xc420057d60 pc=0x40041a4
gocv.io/x/gocv._Cfunc_CascadeClassifier_DetectMultiScale(0xb112c10, 0xb135390, 0x0, 0x0)
	gocv.io/x/gocv/_obj/_cgo_gotypes.go:644 +0x57 fp=0xc420057dd0 sp=0xc420057da0 pc=0x409a9e7
gocv.io/x/gocv.(*CascadeClassifier).DetectMultiScale.func1(0xb112c10, 0xb135390, 0x0, 0x0)
	/Users/Dmitry/go/src/gocv.io/x/gocv/objdetect.go:53 +0xb5 fp=0xc420057e18 sp=0xc420057dd0 pc=0x409c5f5
gocv.io/x/gocv.(*CascadeClassifier).DetectMultiScale(0xc420057f20, 0xb135390, 0x0, 0x0, 0x0)
	/Users/Dmitry/go/src/gocv.io/x/gocv/objdetect.go:53 +0x63 fp=0xc420057e90 sp=0xc420057e18 pc=0x409bcc3
main.main()
	/Users/Dmitry/go/src/github.com/0just0/test_cv/cv.go:49 +0x24b fp=0xc420057f80 sp=0xc420057e90 pc=0x409cd8b
runtime.main()
[...]

Issue #30, opened 2017-11-18
imgproc.cpp issue on MacOS


jkuri:
Hi! I am getting
```
imgproc.cpp:61:28: error: expected '(' for function-style cast or type construction
```
error, tried running simple `Hello World!` sample.

deadprogram:
Hello @jkuri make sure you have set whatever needed environment vars for your config. For example, on OS X or Ubuntu you run `source ./env.sh` as documented here: https://github.com/hybridgroup/gocv#os-x-1

Hope that helps!

jkuri:
Hi Rob, thank you for your support. Now I have
```
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1
```
My clang version is:
```
Apple LLVM version 9.0.0 (clang-900.0.38)
Target: x86_64-apple-darwin17.2.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```
and Xcode `9.1 (9B55)`. Maybe know what's wrong? Thanks again.

deadprogram:
As I see you are using OS X, did you install OpenCV using Homebrew?

jkuri:
yes, I installed it as `brew install opencv3 --with-contrib`. I just also tried to reinstall whole Xcode but still same issue.

jkuri:
I made it work, I updated CGO flags in `env.sh` file to
```sh
export CGO_CPPFLAGS="-I/usr/local/Cellar/opencv/3.3.1_1/include  -I/usr/local/Cellar/opencv/3.3.1/include/opencv2"
export CGO_CXXFLAGS="--std=c++1z -stdlib=libc++"
export CGO_LDFLAGS="-L/usr/local/Cellar/opencv/3.3.1_1/lib -lopencv_core -lopencv_videoio -lopencv_imgproc -lopencv_highgui -lopencv_imgcodecs -lopencv_objdetect -lopencv_calib3d -lopencv_video"
```
path to OpenCV was wrong. Thanks for your support!

deadprogram:
Glad you found that, the `dev` branch already has that change. Seems like we better release it!

jkuri:
yes, or maybe add condition to the `env.sh` for backward compatibility for users who doesn't upgrade libraries very often.

cskonopka:
After following these directions I am a getting the error specified above. Has anyone else had this issue since?

[...]

PR #185, opened 2018-05-06
doc: try and read haarcascade_frontalface_default from GOPATH

adamdecaf:
I tried the face detect example in the readme, but ran into the same
panic as several other users have reported.

I assume many users have pulled down this code on their GOPATH, so
it seems like a viable search path.

See:

- https://github.com/hybridgroup/gocv/issues/25
- https://github.com/hybridgroup/gocv/issues/30
- https://github.com/hybridgroup/gocv/issues/68
- https://github.com/hybridgroup/gocv/issues/72
- https://github.com/hybridgroup/gocv/issues/108


adamdecaf:
If this change is too big what about just the following?

```
	if !classifier.Load(xmlFile) {
		fmt.Printf("Error reading cascade file: %v\n", xmlFile)
		return
	}
```

deadprogram:
I think this has already been handled, or did we miss that change somewhere?

https://github.com/hybridgroup/gocv/blob/master/cmd/faceblur/main.go#L56
https://github.com/hybridgroup/gocv/blob/master/cmd/facedetect/main.go#L59

adamdecaf:
@deadprogram Do those update the readme though? That's where I copied the code from.

deadprogram:
@adamdecaf you are quite correct the README does need update. A PR to correct that would be appreciated!

adamdecaf:
@deadprogram Updated!

Issue #192, opened 2018-05-15
[feature] Get string representation of a video codec


denismakogon:
Assume you have a video file that has the following parameter(s):
 - codec: `H.264`

According to GoCV docs you can get video codec using the following code:
```
video.Get(gocv.VideoCaptureFOURCC)
```
which returns `float64`

There should be built-in method to get a video codec string representation. Unfortunately, there is no built-in feature in OpenCV except manual FOURCC parsing, so i suggest to make it available in GoCV:
```go
package main

import (
	"fmt"
)

func Codec(codecID int64) string {
	res := ""
	hexs := []int64{0xff, 0xff00, 0xff0000, 0xff000000}
	for i, h := range hexs {
		res += string(codecID & h >> (uint(i * 8)))
	}
	return res
}


func main() { 
	c := Codec(int64(828601953))
	fmt.Println(c)
}

```
Link: https://play.golang.org/p/5zmVJffYST9
This function turns FourCC into a sane representation of the codec that can be used later with video writer.

Proposed signature:
```go
func (v *VideoCapture) CodecString() string
```

Implementation:
```go
func (v *VideoCapture) CodecString() string {
	res := ""
	hexs := []int64{0xff, 0xff00, 0xff0000, 0xff000000}
	for i, h := range hexs {
		res += string(int64(v.Get(VideoCaptureFOURCC)) & h >> (uint(i * 8)))
	}
	return res
}
```

deadprogram:
Seems like a very good idea to me!

denismakogon:
closing one as done

deadprogram:
I will close this issue after the next release. Thanks again @denismakogon

PR #193, opened 2018-05-15
Feature: CodecString()

denismakogon:
New API: CodecString() returns a string representation of FourCC bytes from OpenCV video options.

Closes: #192


deadprogram:
Thank you very much @denismakogon now merging.

Issue #196, opened 2018-05-16
Not all parameters to VideoWriter supported


denismakogon:
## Problem
`isColor` is missing and causes problems for H264, AVC1, X264 codecs. According to the [following question](http://answers.opencv.org/question/77893/opencv-3-x264-ffmpeg/), the problem can be solved like this. The problem had nothing to do with the displayed warning. I was trying to write single-channel images, while the VideoWriter was expecting a 3-channel color image (default value of isColor, the 5-th argument to VideoWriter's constructor, is "true"). The solution was setting isColor to false.

But GoCV rely on the default value of `isColor` which is set to `true` by OpenCV as default one.

## Code/Infra
```go
finalVideo, err := gocv.VideoWriterFile(fName, "AVC1", float64(fps), int(width), int(height))
```
Operating system: `apline linux 3.7`
FFMPEG version: `2.8.11-r0`
OpenCV version: `3.4.1`
Go version: `1.10.1`

## Observation
Logs from OpenCV/FFMPEG:
```
OpenCV: FFMPEG: tag 0x31435641/'AVC1' is not supported with codec id 28 and format 'mp4 / MP4 (MPEG-4 Part 14)'
OpenCV: FFMPEG: fallback to use tag 0x00000021/'!???'
```

## Links
[Source](https://docs.opencv.org/3.4.1/dd/d9e/classcv_1_1VideoWriter.html#af52d129e2430c0287654e7d24e3bbcdc) | [Problem](https://github.com/hybridgroup/gocv/blob/master/videoio.cpp#L51) | [Subsequent problem](https://github.com/hybridgroup/gocv/blob/master/videoio.go#L233)

PR #197, opened 2018-05-16
Add isColor to VideoWriterFile

denismakogon:
Add isColor to VideoWriterFile
    
Closes: #196

### DIFF
--- a/videoio.cpp
+++ b/videoio.cpp
@@ -49,7 +49,7 @@ void VideoWriter_Close(VideoWriter vw) {
 }
 
 void VideoWriter_Open(VideoWriter vw, const char* name, const char* codec, double fps, int width,
-                      int height, bool isColor = true) {
+                      int height, bool isColor) {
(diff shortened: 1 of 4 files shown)


deadprogram:
Hi @denismakogon thank you for the PR.

As you already probably know, we have been normally adding 2 versions of any given wrapper in the case of default params: one that just uses all the defaults, and another "WithParams'.

However, in this case, since it is just a single param, I am just wondering if it justfies breaking the API to avoid extra code that is all just for a single bool param. My personal impression is yes.

Before I merge this PR does anyone else want to express an opinion here?

denismakogon:
Actually, i didn’t know that. Thank you for pointing this out. I can update the PR and add one more wrapper to keep API stable, not a problem.

deadprogram:
After some deliberation, I think we should just merge this PR, and go with it. Which I am doing now. Thanks @denismakogon

Issue #195, opened 2018-05-15
gocv.ReadNetFromTensorflow() do not accept config argument


clover113:
https://docs.opencv.org/trunk/d6/d0f/group__dnn.html#gad820b280978d06773234ba6841e77e8d

As shown in this page, the original method in C can accept an extra String config, which is useful when providing pbtxt and pb files separately. When using models from keras, this is essential as the pbtxt must be modified for the model itself to run on OpenCV directly.

https://docs.opencv.org/trunk/d6/d0f/group__dnn.html#ga3b34fe7a29494a6a4295c169a7d32422
Also, a support for cv::dnn::readNet is missing, which can accept the extra config argument.

Proposed method: 
`gocv.readNetFromTensorflowWithConfig(model string, config string) Net {}`
as golang obviously does not support simple overloading and this provides a golang way.
`gocv.readNet(model string, config string, framework string) Net {}`
as readNet method needs to provide a unified interface.

deadprogram:
Hi @clover113 that seems like a very good idea, love to receive a PR with that! :)

clover113:
Yeah, but I do not have experiences with C++ programming or header files. ;-)

clover113:
Ok, I will try forking it.

clover113:
@deadprogram It seems that ReadNet() cannot be implemented as it requires newer versions of OpenCV which even Arch Linux does not provide.

deadprogram:
The next release of GoCV will include `ReadNet()` and will require OpenCV 3.4.2+ so you will likely need to build OpenCV from source.

deadprogram:
The `gocv.ReadNet()` code is now released as part of v0.14.0 so closing this issue. Please reopen if needed. Thank you!

PR #201, opened 2018-05-21
Add ReadNetFromTensorflowWithConfig() method.

clover113:
Add ReadNetFromTensorflowWithConfig() method which allows passing a prototxt aside with the protobuf.
#195

### DIFF
--- a/dnn.cpp
+++ b/dnn.cpp
@@ -10,6 +10,11 @@ Net Net_ReadNetFromTensorflow(const char* model) {
     return n;
 }
 
+Net Net_ReadNetFromTensorflowWithConfig(const char* model, const char* config) {
+	Net n = new cv::dnn::Net(cv::dnn::readNetFromTensorflow(model, config));
+	return n;
+}
+
 void Net_Close(Net net) {
     delete net;
 }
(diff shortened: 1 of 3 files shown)


deadprogram:
Hi @clover113 looks good. Only thing that still is needed is a test to cover the new function. Copying the existing test and adding some code for the new param is all that is needed, nothing facy.

clover113:
@deadprogram I will check if there are existing protobufs and prototxts for testing.

deadprogram:
I think that this can be modified to also extract the file you need for the test: https://github.com/hybridgroup/gocv/blob/master/travis_build_opencv.sh#L70

You will also need to modify https://github.com/hybridgroup/gocv/blob/master/appveyor_build_opencv.cmd#L20

deadprogram:
Hello @clover113 there is some work in progress in the https://github.com/hybridgroup/gocv/tree/feature/openvino-ie-support branch that applies to your needs. The newly implemented `gocv.ReadNet()` function allows you to use a config file aka `.prototxt` and also supports all of the DNN frameworks that are part of OpenCV's backends, which includes TensorFlow.

This might be a better solution in the sense that is provides what you need, without having to create yet another function.

What do you think?

Please note that this branch requires the upcoming OpenCV 3.4.2, which should be released very soon.

deadprogram:
This PR is no longer needed thanks to the `ReadNet()` function. Now closing.

PR #206, opened 2018-05-25
Use gocv.VideoCaptureFile if string is specified for device.

mattn:
gocv.VideoCaptureFile is useful to use video file instead of webcam. And also, some of devices (ex: jetson tx2) doesn't have v4l2 compatible devices on Linux. On jetson tx2, gstreamer query can be used for this device name instead of device ID.


mattn:
For example, below is a command that I could facedetect on jetson tx2.

```
$ ./main "nvcamerasrc ! video/x-raw(memory:NVMM), format=NV12 ! nvvidconv flip-method=0 ! videoconvert ! video/x-raw, format=BGR ! appsink" haarcascade_facedetect.xml
```

deadprogram:
Hi @mattn this is good stuff. The only possible drawback is that now each cmd example is no longer just a single, standalone file. Is that really all that important for the learner? Perhaps those concerns are best solved with tutorials.

The only other thought I had, was perhaps we need to add your code as `gocv.OpenVideoCapture()` to `videoio.go`, since it is such a very common case.

What do you think? Any opinions anyone?

mattn:
Good idea. I'll send PR in later.

deadprogram:
Any chance of getting a PR soon? Thanks!

mattn:
os.Args[0] is always string so the function you mentioned should call VideoCaptureFile every time.

deadprogram:
Even though we need some test coverage for the new code, I am thinking I will merge to `dev` anyhow. Test coverage can be added subsequently.

Thank you again for working on this, @mattn

PR #210, opened 2018-05-26
Support pkg-config via cgo directives

### LINKED ISSUE #126: Support pkg-config and/or external library
Currently I don't see LDFLAGS defined anywhere in the code, I had to export CGO_LDFLAGS in order to `go get` this package. 

```
export CGO_LDFLAGS="-lopencv_core -lopencv_videoio -lopencv_imgproc -lopencv_imgcodecs -lopencv_objdetect -lopencv_video -lopencv_highgui -lopencv_dnn -lopencv_features2d"
```
Instead of just instructions how to build OpenCV in specific way why not allow something like this `//#cgo pkg-config: opencv` , to be used with `-tags pkg-config`.  Or/and add all required libraries to LDFLAGS so it can be used with `-tags extlib` or something like that.

My pkg-config for example returns correct flags and gocv can be built with those:
```
# pkg-config --cflags --libs opencv
-I/usr/include/opencv -lopencv_dnn -lopencv_ml -lopencv_objdetect -lopencv_shape -lopencv_stitching -lopencv_superres -lopencv_videostab -lopencv_calib3d -lopencv_features2d -lopencv_highgui -lopencv_videoio -lopencv_imgcodecs -lopencv_video -lopencv_photo -lopencv_imgproc -lopencv_flann -lopencv_core
```

vcabbage:
Tested on macOS and in a Linux docker container. I haven't tested on Windows but copied the values from `env.cmd`.

Renamed `highgui.h` to `highgui_gocv.h` because pkg-config adds `-I/usr/local/include/opencv`, which also contains a `highgui.h`. The `highgui.h` in `/usr/local/include/opencv` ends up being included instead of the local one. If there's a better way to avoid the conflict I'm happy to update the PR.

Resolves #126


deadprogram:
This is a great convenience for people using the simple defaults on most OS. Thanks for adding it, @vcabbage 

Perhaps you can also update the README to match, and also a note on how to use `-tags customenv` when needed for unique builds.

@maruel any thoughts on this? You are the build expert :building_construction:

cgo.go
```diff
@@ -0,0 +1,11 @@
+// +build !customenv
+
+package gocv
+
+/*
+#cgo !windows pkg-config: opencv
+#cgo darwin   CXXFLAGS:   --std=c++1z -stdlib=libc++
+#cgo windows  CPPFLAGS:   -IC:\opencv\build\install\include
```
maruel: Hardcoding the paths here is a bit user-hostile IMHO and would be really tricky to debug for users.

cgo.go
```diff
@@ -0,0 +1,11 @@
...
+#cgo windows  LDFLAGS:    -LC:\opencv\build\install\x64\mingw\lib -lopencv_core341 -lopencv_face341 -lopencv_videoio341 -lopencv_imgproc341 -lopencv_highgui341 -lopencv_imgcodecs341 -lopencv_objdetect341 -lopencv_features2d341 -lopencv_video341 -lopencv_dnn341 -lopencv_xfeatures2d341 -lopencv_plot341 -lopencv_tracking341 -lopencv_img_hash341
```
maruel: That fact that the version number is now hardcoded here makes the upgrade process much more opaque. I can certainly imagine someone rolling the default opencv version in the makefile, pushing that without testing on Windows and forgetting to update these values.

cgo.go
```diff
@@ -0,0 +1,11 @@
+// +build !customenv
+
+package gocv
+
+/*
+#cgo !windows pkg-config: opencv
+#cgo darwin   CXXFLAGS:   --std=c++1z -stdlib=libc++
```
maruel: I don't think OpenCV requires C++17?
Why enforce the stlib used on macOS?

vcabbage: I copied the values from `env.sh`. Just tested and it builds with just `--std=c++11`.

cgo.go
vcabbage: The Windows flags were copied directly from `env.cmd` I didn't notice the hardcoded version number.

Do you know if there's a better way to go about this for Windows? Windows could be omitted, but to me that seems less friendly to users than providing some defaults that can be overridden.

Hopefully AppVeyor would catch a mismatch with the default version. Probably a good idea to remove explicitly setting the `CGO_*` variables in `appveyor.yaml` and `.travis.yml`

vcabbage:
@deadprogram Sure thing on the README. Should I remove `env.sh` and `env.cmd` or do you want to keep those?

cgo.go
deadprogram: That is what the `env.cmd` is doing now, so we're not any worse off with this change. If you wish to build using a custom path, that would be something we can document how to do.

cgo.go
deadprogram: That might have been an artifact I left in which trying to work on macOS builds at some point. Seems familiar. IN any case, best to only enforce the minimum needed version, right?

deadprogram:
We should probably replace the env.sh and env.cmd script implementations to just display a deprecation warning and refer the user to the new README section on setting up a custom build environment for when that is needed, for example in the case of OpenVINO.

cgo.go
maruel: I mean it's worth off in the sense that if someone upgrades their opencv version they do not have to modify the sources to be able to build. But I don't really mind personally.

cgo.go
maruel: Too bad.

cgo.go
maruel: It probably doesn't matter in practice, but I generally think the less flags the better. :)

vcabbage:
Updated the flags and removed the vars from the CI config. Tested on Windows and found I needed to adjust the directives a bit.

`#cgo windows  CXXFLAGS:   --std=c++11` seems like it could be a problem since I saw the same flag cause the older version of gcc on Travis to fail, but it was required for me to compile on my Windows system. If it does come up it can be worked around by updating the gcc (I assume) or using the customenv tag. Is that acceptable?

Still working on the README updates.

vcabbage:
Updated the scripts and README. Should be ready for another review.

deadprogram:
This is really useful work, @vcabbage thank you very much. Now merging into `dev` branch.

maruel:
Thanks, the change is very good. The only nit I'd say is to add a comment in the Makefile around the OPENCV_VERSION variable definition at the top to put a recall about the need to update the corresponding Go source file, otherwise people will forget. Other than that it's a great improvement.

PR #219, opened 2018-06-04
readme: fix build of face tracking sample

mewmew:


deadprogram:
Thanks again @mewmew now merging.

PR #234, opened 2018-06-13
Feature/image to mat

edwvee:
First of all:
Fixed ToImage. NRGBA -> RGBA. With RGBA alpha is darker than it is. Actually, I wrote my functions converting image to mat at first and test. But all was coming wrong when there was alpha channel, because of implementation of Mat's to Image. Test with RGBA didn't pass. And, as an experiment, I converted image with ImageToMatRGBA to Mat and back with Mat's ToImage to image and saved as png. Transparent zones became opaquer. That looks strange, because I actually just took RGBA, converted to 8bit per channel, put it to slice as BGRA and returned it back to image.
I guess, opencv doesn't use a non-alpha-premultiplied RGBA version.
And, finally there are functions for converting and tests for them.


deadprogram:
Hi @edwvee thank you very much for working on this PR.

Could you please also add a test to cover the `ImageToMatGray()` function?

edwvee:
Well, build failed. But, looks live it has nothing to do with my commits. Moreover, I've run failed test and it passed.

edwvee:
TestCaffe and:
```
Error downloading remote file: One or more errors occurred.
Inner Exception: The remote name could not be resolved: 'dl.caffe.berkeleyvision.org'
```

deadprogram:
Seems like Caffe download server DNS is currently unreachable. :/

core_test.go
```diff
@@ -1182,6 +1183,33 @@ func TestImageToMatRGB(t *testing.T) {
...
+	}
+	defer file.Close()
+	imgSrc, _, err := image.Decode(file)
+	if err != nil {
+		log.Fatal(err)
+	}
+	img0 := image.NewGray(imgSrc.Bounds())
+	draw.Draw(img0, imgSrc.Bounds(), imgSrc, image.ZP, draw.Src)
+
+	mat, err := ImageToMatRGB(img0)
```
deadprogram: Shouldn't this be `ImageToMatGray()` here?

edwvee: Facepalm to me =)

core.go
```diff
@@ -500,13 +500,75 @@ func (m *Mat) ToImage() (image.Image, error) {
...
+	data := make([]byte, 0, x*y)
+	switch img.(type) {
+	case *image.Gray:
+		imgG := img.(*image.Gray)
+		for j := bounds.Min.Y; j < bounds.Max.Y; j++ {
+			for i := bounds.Min.X; i < bounds.Max.X; i++ {
+				data = append(data, imgG.GrayAt(i, j).Y)
+			}
+		}
+	default:
```
deadprogram: The following code is intended to convert RGB to grayscale? Not sure if this is the best way to accomplish this, but if so it also needs test coverage.

edwvee: I had thoughts about it. There are different formula's. I think it's better to leave for user which way to choose.

edwvee:
Also, there must be a change of name to not confuse user.

deadprogram:
Thanks very much for working on this @edwvee now merging.

PR #241, opened 2018-06-22
Added opencv::calcHist implementation

milosgajdos83:
This adds `CalcHist` function which brings in support for [calcHist](https://docs.opencv.org/3.4.1/d6/dc7/group__imgproc__hist.html#ga6ca1876785483836f72a77ced8ea759a) opencv function to gocv.

I need to add test to it -- I've verified the implementation on some basic examples in one of my personal projects, but I need to come up with a few simple test cases for `gocv`.

### DIFF
--- a/core.h
+++ b/core.h
@@ -20,6 +20,12 @@ typedef struct IntVector {
     int length;
 } IntVector;
 
+// Wrapper for std::vector<float>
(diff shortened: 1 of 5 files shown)


deadprogram:
This is great, thank you very much @milosgajdos83 now merging.

PR #251, opened 2018-07-11
Add Mat.DataPtr methods for direct access to OpenCV data

### LINKED ISSUE #250: Direct access to Mat data
## Description

I have a case where I need access to the underlying bytes of a Mat, but don't hold on to them longer than the Mat exists. As an optimization I access the C allocated data directly rather than copying it into Go memory with `Mat.ToBytes`. I'm currently using `Mat.Ptr` to accomplish this, but it's rather ugly.

Would it make sense to add a method which allows this more easily? From what I've seen, the existing API strives to be safe, so I'm not sure how well it would fit with the goals of this project. Perhaps something like `func (*Mat) DataPtr() unsafe.Pointer` to reduce the chance of it being used inadvertently. In my case I'm dealing with the CV_32FC2 result from `CalcOpticalFlowFarneback`, so I have to use unsafe to cast to a `[]float32` anyway.

Thanks for the project and taking the time to consider this issue.

## Steps to Reproduce

n/a

## Your Environment

n/a

vcabbage:
@deadprogram It doesn't seem necessary to return an error for the single byte variants. The int8 version probably isn't that useful, but I could imagine a scenario where someone wants to access the raw bytes even if the Mat is of a different type.

Resolves #250

### DIFF
--- a/core.cpp
+++ b/core.cpp
@@ -66,6 +66,10 @@ struct ByteArray Mat_ToBytes(Mat m) {
     return toByteArray(reinterpret_cast<const char*>(m->data), m->total() * m->elemSize());
 }
 
+struct DataPtr Mat_DataPtr(Mat m) {
+    return DataPtr {reinterpret_cast<const char*>(m->data), m->total() * m->elemSize()};
+}
+
 // Mat_Region returns a Mat of a region of another Mat
 Mat Mat_Region(Mat m, Rect r) {
     return new cv::Mat(*m, cv::Rect(r.x, r.y, r.width, r.height));
--- a/core.cpp
+++ b/core.cpp
@@ -66,8 +66,8 @@ struct ByteArray Mat_ToBytes(Mat m) {
     return toByteArray(reinterpret_cast<const char*>(m->data), m->total() * m->elemSize());
 }
 
-struct DataPtr Mat_DataPtr(Mat m) {
-    return DataPtr {reinterpret_cast<const char*>(m->data), m->total() * m->elemSize()};
+struct ByteArray Mat_DataPtr(Mat m) {
+    return ByteArray {reinterpret_cast<char*>(m->data), static_cast<int>(m->total() * m->elemSize())};
 }
 
 // Mat_Region returns a Mat of a region of another Mat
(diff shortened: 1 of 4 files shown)


core_test.go
```diff
@@ -213,6 +213,99 @@ func TestMatToBytes(t *testing.T) {
 	}
 }
 
+func TestMatDataPtr(t *testing.T) {
```
deadprogram: Seems to me like it would be a good idea to have code in these tests to set/get a few values in the Mat.

vcabbage:
Added the get/set tests, which made me realize that I was using `Mat_ToBytes` improperly since it allocates and copies the bytes in C++. I added a new `Mat_DataPtr` function to avoid that. I don't normally write C++, so I'd appreciate a double check of that code/any suggestions to make it better.

I think `Mat_DataPtr` could be used for `Mat.ToBytes` as well so that the data data isn't copied twice (once in C++ and once in Go), but I didn't want to expand the scope of this change.

core.cpp
```diff
@@ -66,6 +66,10 @@ struct ByteArray Mat_ToBytes(Mat m) {
     return toByteArray(reinterpret_cast<const char*>(m->data), m->total() * m->elemSize());
 }
 
+struct DataPtr Mat_DataPtr(Mat m) {
```
deadprogram: Seems like we could reuse ByteArray struct here instead of creating a new type. Basically just do what you already implemented but with ByteArray, and since Mat_DataPtr does not perform the memcpy, it does the right thing.

core.h
```diff
@@ -15,6 +15,11 @@ typedef struct ByteArray {
     int length;
 } ByteArray;
 
+typedef struct DataPtr {
```
deadprogram: See previous comment, seems like could just use the ByteArray type instead of adding an additional type.

core_test.go
deadprogram: Looking good now, thanks for adding the additional tests.

vcabbage:
Removed the `DataPtr` struct and replaced `Mat_ToBytes` with `Mat_DataPtr` in `Mat.ToBytes`.

deadprogram:
@vcabbage this is really good stuff. Seems ready to merge now, right?

vcabbage:
Yep, should be all set. Thanks for reviewing!

PR #254, opened 2018-07-12
MinAreaRect()

covrom:
MinAreaRect finds a rotated rectangle of the minimum area enclosing the input 2D point set.


deadprogram:
Thank you for the PR @covrom now merging.

PR #281, opened 2018-08-24
Improved download and archive extraction speed during setup

viktor-ferenczi:
Disabled progress bar for improved download and archive extraction performance.
On fast internet and SSD these changes can save more than two minutes.

Affects: 
- Invoke-WebRequest
- Expand-Archive

Also added extra comments to explain to the user in case of slower Internet connectivity.


deadprogram:
Those are some great benchmarks, @viktor-ferenczi thanks for working on this. That is some very key research there to find https://github.com/PowerShell/PowerShell/issues/2138

I switched the branch for your PR to `dev` and will now merge. Thanks again.

Issue #282, opened 2018-08-25
'make install' endless loop on CentOS 7


justinclift:
On CentOS 7 x64, `make install` goes into an endless loop.

It looks like the Makefile is expecting a (probably newer) version of wget than provided with CentOS:

```
wget: unrecognized option '--show-progress'
Usage: wget [OPTION]... [URL]...

Try `wget --help' for more options.
unzip:  cannot find or open opencv.zip, opencv.zip.zip or opencv.zip.ZIP.
wget: unrecognized option '--show-progress'
Usage: wget [OPTION]... [URL]...

Try `wget --help' for more options.
unzip:  cannot find or open opencv_contrib.zip, opencv_contrib.zip.zip or opencv_contrib.zip.ZIP.
rm: cannot remove ‘opencv.zip’: No such file or directory
rm: cannot remove ‘opencv_contrib.zip’: No such file or directory
```

Removing the `--show-progress` option from the wget line seems to work, although the script bombs out later anyway with `-- Configuring incomplete, errors occurred!`.  (going to look into that shortly)

As a general thought for troubleshooting, it'd probably be useful to have the Makefile detect when something goes wrong so it can skip the automatic `rm -rf` of the temp directory it builds OpenCV in. :wink:

## Your Environment

* Operating System and version:
  * CentOS 7 x64.  Specifically `CentOS Linux release 7.5.1804 (Core)`.
* OpenCV version used:
  * 3.4.2
* How did you install OpenCV?
  * `go get -u -d gocv.io/x/gocv`
* GoCV version used:
  * **3.4.2** is being attempted
* Go version:
  * 1.11
* Did you run the `env.sh` or `env.cmd` script before trying to `go run` or `go build`?
  * No, there's no mention of them in the initial docs.  Should there be?

deadprogram:
Your recent PR was merged into `dev` branch @justinclift I will close this issue once it has been released. Thanks again.

deadprogram:
This has been released as part of v0.16.0 thank you.

PR #283, opened 2018-08-25
Fix package names for CentOS 7 and Fedora 28

justinclift:
The present Makefile has a typo in the (RPMS) package name, for CentOS 7 and Fedora 28.

This PR fixes that, and also addresses the wget problem in #282. :smile:

It also adds `unzip` to the package list for CentOS and Fedora, as that's needed and isn't part of the general base system.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ OPENCV_VERSION?=3.4.2
 TMP_DIR?=/tmp/
 
 # Package list for each well-known Linux distribution
-RPMS=cmake git gtk2-devel libpng-devel libjpeg-devel libtiff-devel tbb tbb-devel libdc1394-devel
+RPMS=cmake git gtk2-devel libpng-devel libjpeg-devel libtiff-devel tbb tbb-devel libdc1394-devel unzip
 DEBS=unzip build-essential cmake git libgtk2.0-dev pkg-config libavcodec-dev libavformat-dev libswscale-dev libtbb2 libtbb-dev libjpeg-dev libpng-dev libtiff-dev libdc1394-22-dev
 
 # Detect Linux distribution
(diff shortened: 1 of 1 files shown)


justinclift:
Just checked on Fedora 28... and the package name there is `pkgconf-pkg-config` instead.

So, this PR needs adjusting.  It'll address that shortly.

justinclift:
Btw, tested the changes on both Fedora 28 and CentOS 7.  These work as intended on both.

CentOS 7 still needs further work, but it's not stuff related to this. :smile:

justinclift:
Hmmm, it'd probably be better to add both `unzip` and `wget` to the package list too, as neither of those are part of the minimal build.  On the other hand, it should be `curl` instead of `wget` anyway, as `curl` is part of the base system.

Just added a commit for that.  Someone using Debian/Ubuntu will probably want to check the `curl` command on their system though, just to see if it's part of the base install.  I'm not sure. :wink:

deadprogram:
Thanks for all this great stuff, @justinclift I will test the Debian stuff and get back to you shortly.

deadprogram:
OK, confirmed that it works on Ubuntu. This is terrific, thanks again for working on it @justinclift now merging.

justinclift:
No worries at all @deadprogram. :smile:

justinclift:
As a general observation, on CentOS 7 the standard gcc available (4.8.x) doesn't do enough C++11.  So, the `cmake` fails.

The workaround I used was to install the (official) "Developer Toolset 7" from the (Red Hat backed) "Software Collections" project:

&nbsp; &nbsp; https://www.softwarecollections.org/en/scls/rhscl/devtoolset-7/

That supplies a newer gcc among other things, so the `cmake` passes and things work.

I'm not sure if that's how everyone else using OpenCV on CentOS 7 is doing it though, so I haven't turned that into docs or steps for people to follow (which I often do).  On that note, there's no wiki here.  Is there some place to write down steps for other people to leverage, or is that more the kind of thing that goes into the book you're doing. :smile:

deadprogram:
At the moment that kind of info has just been getting put into the README and then the gocv.io web site pages. A PR adding it seems like a good idea and would be appreciated.

justinclift:
Oh, as a side thought... do you reckon turning on the wiki would be useful for people?

deadprogram:
I thought about it, but I think at the moment we are better off updating the web site via PRs at https://github.com/hybridgroup/gocv-site

justinclift:
No worries at all. :smile:

PR #297, opened 2018-09-08
add condition to handle no circle found in circle detection example

zak905:
When the image has no circles, the following error is thrown: `runtime error: index out of range` this is because v in https://github.com/hybridgroup/gocv/blob/master/cmd/find-circles/main.go#L62 has length 1 or 0. Adding a lenght check fixes the issue

### DIFF
--- a/cmd/find-circles/main.go
+++ b/cmd/find-circles/main.go
@@ -60,12 +60,15 @@ func main() {
 
 	for i := 0; i < circles.Cols(); i++ {
 		v := circles.GetVecfAt(0, i)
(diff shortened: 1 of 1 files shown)


deadprogram:
Hello @zak905 thank you for the PR. I changed it to the `dev` branch as discussed here: https://github.com/hybridgroup/gocv/blob/master/CONTRIBUTING.md#how-to-use-our-github-repository

Thanks again, now merging.

zak905:
Your welcome, I am aware of that :) I forgot to check which branch

PR #298, opened 2018-09-10
Added Phase function

dlion:
Hi,

I added the `phase` function to `core` and updated the `ROADMAP`

Regards


deadprogram:
That is great @dlion thank you very much! Now merging.

PR #308, opened 2018-09-30
Mat profile

dougnd:
I had previously asked whether there was anyway to close gocv.Mats automatically.  The response made sense, but I have found tracking memory leaks do to forgetting to close gocv.Mats very annoying.  Then I learned about pprof.Profiles.  They make tracking resource leaks a breeze.   My suggestion is that you make it part of gocv.   I have coded an initial implementation and showed how it can be used to track unclosed images in tests.  Let me know what you think, I can modify this as you see fit.

To give you a better idea of this, here is the proposed godoc string for the profile:
MatProfile a pprof.Profile that contains stack traces that led to (currently) unclosed Mat's creations.  Every time a Mat is created, the stack trace is added to this profile and every time the Mat is closed the trace is removed. In a program that is not leaking, this profile's count should not continuously increase and ideally when a program is terminated the count should be zero.  You can get the count at any time with:

    gocv.MatProfile.Count()

and you can display the current entries with:

    var b bytes.Buffer
    gocv.MatProfile.WriteTo(&b, 1)
    fmt.Print(b.String())

This will display stack traces of where the unclosed Mats were instantiated. For example, the results could look something like this:

    1 @ 0x4146a0c 0x4146a57 0x4119666 0x40bb18f 0x405a841
    #	0x4146a0b	gocv.io/x/gocv.newMat+0x4b	/go/src/gocv.io/x/gocv/core.go:120
    #	0x4146a56	gocv.io/x/gocv.NewMat+0x26	/go/src/gocv.io/x/gocv/core.go:126
    #	0x4119665	gocv.io/x/gocv.TestMat+0x25	/go/src/gocv.io/x/gocv/core_test.go:29
    #	0x40bb18e	testing.tRunner+0xbe		/usr/local/Cellar/go/1.11/libexec/src/testing/testing.go:827

Furthermore, if the program is a long running process or if gocv is being used on a web server, it may be helpful to install the HTTP interface using:

    import _ "net/http/pprof"

**Important point about the diff**:  One change you should investigate thoroughly is the removal of line 42 from dnn_ext.go.  Calling `Region` on a Mat will allocate another Mat and return it, but in this case the return value is not used.  I'm not sure if there is a bug there or if that line was just unneeded. I didn't investigate fully so I'm not sure removing it is what I should have done.


deadprogram:
I am not very well versed in pprof @dougnd so it is taking me a bit to understand this. Thank you for working on it, and I will be providing more feedback very soon.

deadprogram:
Sorry it took so long to get back to this most excellent PR, @dougnd thank you again for working on it. Seems like the only thing we need now is to add some instructions to the README on how to use it! Do you mind adding a new section for "Profiling"?

dougnd:
I added some documentation to the README.  Let me know if I should add more (or reduce it!). 

Did you take a look at line 42 of dnn_ext.go?  I had to remove it to satisfy the new leak checking in the tests since the line allocates a `Mat`.

The line is:
```
    square.Region(roi)
```

Since the result of the Region is never used, that line of code seems ineffective (and thus I removed it).  I don't know enough about that section of the code to know if removing it is correct though.  Perhaps it is supposed to actually be doing something?

deadprogram:
Hi @dougnd I have now merged this PR but with a few changes. The main one is that there is now a  build tag `matprofile` to control if the custom profiler is used or not. If the tag is not there, the profiler is not used.

So to run the mjpeg-server with the MatProfile turned on, you would run:

```
go run -tags matprofile ./cmd/mjpeg-streamer/main.go 0 0.0.0.0:8080
```

Thanks again for this great effort, it will really come in handy.

Issue #310, opened 2018-10-01
Add Sort operation wrapper to Arrays 


morfeo8marc:
Hi I will start working on adding sort operation.

deadprogram:
Released as part of 0.17. Now closing.

PR #311, opened 2018-10-01
core: add Sort() method to Mat

morfeo8marc:
Add sort method for enhancement request #310.


deadprogram:
Looks great @morfeo8marc thank you very much for another fine contribution. Now merging.

Issue #312, opened 2018-10-01
Add Trace operation to Arrays


morfeo8marc:
Hi, I will start working on adding Trace operation to Arrays

deadprogram:
Released as part of 0.17. Now closing.

PR #313, opened 2018-10-01
core: add Trace() method to Mat

morfeo8marc:
Add Trace method for enhancement request #312.

### DIFF
--- a/core.cpp
+++ b/core.cpp
@@ -550,6 +550,16 @@ void Mat_Subtract(Mat src1, Mat src2, Mat dst) {
     cv::subtract(*src1, *src2, *dst);
 }
 
+Scalar Mat_Trace(Mat src) {
+    cv::Scalar c = cv::trace(*src);
+    Scalar scal = Scalar();
+    scal.val1 = c.val[0];
+    scal.val2 = c.val[1];
+    scal.val3 = c.val[2];
+    scal.val4 = c.val[3];
+    return scal;
+}
+
 void Mat_Transform(Mat src, Mat dst, Mat tm) {
     cv::transform(*src, *dst, *tm);
 }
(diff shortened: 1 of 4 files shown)


core.cpp
```diff
@@ -550,6 +550,16 @@ void Mat_Subtract(Mat src1, Mat src2, Mat dst) {
     cv::subtract(*src1, *src2, *dst);
 }
 
+Scalar Mat_Trace(Mat src) {
+    cv::Scalar c = cv::trace(*src);
+    Scalar scal = Scalar();
```
deadprogram: Seems like this could be `Scalar scal = Scalar(c.val[0], c.val[1], c.val[2], c.val[3]);` and eliminate the next 4 lines. What do you think?

morfeo8marc: If tried to do that, but I get the following error:
```
core.cpp: In function ‘Scalar Mat_Trace(Mat)’:
core.cpp:555:64: error: no matching function for call to ‘Scalar::Scalar(double&, double&, double&, double&)’
     Scalar scal = Scalar(c.val[0], c.val[1], c.val[2], c.val[3]);
                                                                ^
In file included from core.cpp:1:0:
core.h:92:16: note: candidate: Scalar::Scalar()
 typedef struct Scalar {
                ^
core.h:92:16: note:   candidate expects 0 arguments, 4 provided
core.h:92:16: note: candidate: constexpr Scalar::Scalar(const Scalar&)
core.h:92:16: note:   candidate expects 1 argument, 4 provided
core.h:92:16: note: candidate: constexpr Scalar::Scalar(Scalar&&)
core.h:92:16: note:   candidate expects 1 argument, 4 provided
```
Do I need to add a "constructor" to `typedef struct Scalar` in `core.h`

deadprogram: My mistake, I did not realize that was the GoCV wrapper `Scalar` not the OpenCV `cv:Scalar` class. Never mind...

deadprogram:
Thank you for the PR @morfeo8marc now merging.

Issue #315, opened 2018-10-02
Add SortIdx operation to Arrays


morfeo8marc:
Hi, I will start working on adding SortIdx operation to Arrays

deadprogram:
Released as part of 0.17. Now closing.

PR #316, opened 2018-10-02
Add SortIdx operation to Arrays

morfeo8marc:
Add SortIdx method for enhancement request #315.


morfeo8marc:
Wrong target branch

Issue #318, opened 2018-10-02
Add Reduce operation to Arrays


morfeo8marc:
Hi, I will start working on adding Reduce Operation to Mat

deadprogram:
Released as part of 0.17. Now closing.

PR #321, opened 2018-10-03
core: add Reduce() method to Mat

morfeo8marc:
Add Reduce method for enhancement request #318.


deadprogram:
Looks great @morfeo8marc thank you very much! Now merging.

Issue #326, opened 2018-10-03
Add SolveCubic operation to Arrays


morfeo8marc:
Hi, I will start working on adding SolveCubic operation to Mat

deadprogram:
Released as part of 0.17. Now closing.

PR #327, opened 2018-10-03
core: add SolveCubic() method to Mat

morfeo8marc:
Add SolveCubic method for enhancement request #326.


deadprogram:
Thank you @morfeo8marc looking good! Now merging.

PR #329, opened 2018-10-04
Added fitLine function

dlion:
Hi, I added the `fitLine` function. 👾


deadprogram:
:triangular_ruler: thank you very much @dlion now merging.

Issue #331, opened 2018-10-05
Add SolvePoly operation to Array


morfeo8marc:
H,
Will start working on adding SolvePoly method.

deadprogram:
Released as part of 0.17. Now closing.

PR #332, opened 2018-10-05
core: add SolvePoly() method to Mat

morfeo8marc:
Add SolvePoly method for enhancement request #331.


deadprogram:
Looks good @morfeo8marc thanks for fixing the test. Now merging.

PR #333, opened 2018-10-05
Update Makefile

wangyufeng0615:
You can repeat "make install" if the make download process is interrupted.


deadprogram:
That certainly makes sense to me. Now merging, thanks @wangyufeng0615

Issue #335, opened 2018-10-07
Add Repeat operation to Arrays


morfeo8marc:
Hi,
I will start working on adding Repeats operations to Arrays

deadprogram:
Released as part of 0.17. Now closing.

PR #336, opened 2018-10-07
core: add Repeat() method to Mat

morfeo8marc:
Add Repeat method for enhancement request #335.


deadprogram:
Thank you for another contribution @morfeo8marc now merging.

PR #357, opened 2018-11-13
Return the right type and error when opening VideoCapture fails

milosgajdos83:
When opening a video capture (file or device) `VideoCapture::open(...)` returns `bool` value depending on success of the `open` call. However we are ignoring the return value and not setting the error when returning from Go bindings functions. This leads to bugs in Go programs which are checking for `error` values returned from `VideoCapture*` functions when the returned error is `nil`, which leads to the code assuming the video capture `open` was successful.

Furthermore `VideoCapture::open(...)` calls return `bool` values. We are however returning integer, which is ok in C/C++ world as it's implicitly cast by `cgo` into particular bool value. However, we want to make this explicit in GoCV so we are returning bool now as per OpenCV.

### DIFF
--- a/videoio.cpp
+++ b/videoio.cpp
@@ -9,11 +9,11 @@ void VideoCapture_Close(VideoCapture v) {
     delete v;
 }
 
-int VideoCapture_Open(VideoCapture v, const char* uri) {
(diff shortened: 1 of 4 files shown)


deadprogram:
Thanks, @milosgajdos83 we needed this. Now merging.

PR #359, opened 2018-11-19
Removing linking against face contrib module

milosgajdos83:
This PR removes `-lopencv_face` from `env.sh` to skip linking openvino binaries with opencv's face contrib module.


deadprogram:
:+1: now merging.

PR #370, opened 2018-12-04
Implement dnn methods for loading in-memory models.

jheidel:
This PR adds missing methods for loading of in-memory deep neural network models. The existing methods only support loading from the filesystem.

### DIFF
--- a/dnn.cpp
+++ b/dnn.cpp
@@ -5,13 +5,31 @@ Net Net_ReadNet(const char* model, const char* config) {
     return n;
 }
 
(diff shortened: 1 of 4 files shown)


deadprogram:
Hi @jheidel sorry it has taken so long to review this PR. Very interesting use case you must have.

Now merging, thank you for your contribution.

jheidel:
Thanks for the review! Just to provide some background on the use case: I was trying to load a model from embedded resources so that my go binary is nicely portable rather than depending on the location of the external model files.

PR #371, opened 2018-12-05
Adds Dockerfile. Updates Makefile and README.

milosgajdos83:
This commit adds a `Dockerfile` which allows to build `GoCV` Docker images so you can build `GoCV` programs that can be run in Docker containers.

A new `make` target, called `docker` was added to the project `Makefile` to make building `GoCV` Docker image easily. Finally, the `Makefile` now correctly sets `cmake` BUILD parameters, too.

Lastly, `README` has been updated with the information about how to build `GoCV` Docker images.


deadprogram:
This is awesome! Now merged.

PR #382, opened 2019-01-01
Fixing filename in caffe-classifier example

stefanoj3:
The caffe-classifier example uses the wrong filename, which might be a bit confusing.

This PR fixes it.

### DIFF
--- a/cmd/caffe-classifier/main.go
+++ b/cmd/caffe-classifier/main.go
@@ -14,11 +14,11 @@
 //
 // How to run:
 //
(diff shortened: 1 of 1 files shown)


deadprogram:
Hi @stefanoj3 thank you for the PR. I switched the branch for it to `dev` as discussed here: https://github.com/hybridgroup/gocv/blob/master/CONTRIBUTING.md#how-to-use-our-github-repository

Now merging. Thanks again!

Issue #342, opened 2018-10-19
minEnclosingCircle and pointPolygonTest


ibenben:
I want to use minEnclosingCircle and pointPolygonTest.How can I do that?

ibenben:
I knew that.Is any other way can help me to use those function?

deadprogram:
They will need to be added to the GoCV wrappers to be used from Go.

ibenben:
@deadprogram Thanks.

deadprogram:
`minEnclosingCircle` was previously added to GoCV however `pointPolygonTest` still needs implementation.

Garykom:
> however `pointPolygonTest` still needs implementation.
https://github.com/hybridgroup/gocv/compare/release...Garykom:release

PR #401, opened 2019-02-07
Add opencv's 'MinEnclosingCircle' function to gocv (and test).

jonathankarsh:
Addressing part of #342

### DIFF
--- a/core.h
+++ b/core.h
@@ -38,6 +38,12 @@ typedef struct RawData {
     struct ByteArray data;
 } RawData;
 
+// Wrapper for an individual cv::Point2f
+typedef struct Point2f {
+    float x;
+    float y;
+} Point2f;
+
 // Wrapper for an individual cv::cvPoint
 typedef struct Point {
     int x;
--- a/imgproc.cpp
+++ b/imgproc.cpp
@@ -189,6 +189,19 @@ struct RotatedRect MinAreaRect(Points points){
     return retrect;
 }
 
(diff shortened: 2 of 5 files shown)


jonathankarsh:
Notes:
1. Not sure about the new struct.
2. Perhaps a "global" const epsilon for the test isn't desired.
3. It seems that minEnclosingCircle is pretty unique in getting parameters by &ref, maybe that warrants a comment or different handling.
4. Forgot to add comment for new method, will do after first review / sanity check.

deadprogram:
Hello @jonathankarsh thank you for the PR.

No need for the new struct, you can just use this signature instead 

```go
func MinEnclosingCircle(points []image.Point) (x, y, radius float32)
```

Also agree, please move the `epsilon` into the body of the test itself as a `var`

`cv::minEnclosingCircle()` is pretty typical of functions in OpenCV (and C++ in general) that return more than one value, generally do so by using params passed by pointer reference.

Yes, please add the GoDoc formatted comment like the other GoCV functions to the new function https://github.com/hybridgroup/gocv/pull/401/files#diff-50cee154353ca3af045f08401c7d081bR339

Lastly, should this type be defined in `core.go` possibly?
https://github.com/hybridgroup/gocv/pull/401/files#diff-50cee154353ca3af045f08401c7d081bR329

Thanks!

jonathankarsh:
Thanks! PTAL

deadprogram:
Looking good now, thanks for making those changes @jonathankarsh 

Now merging, thank you again for contributing.

Issue #389, opened 2019-01-24
Tests failing on macOS


Sauraus:
`make test` on dev & master branches both fails because of a missing file `nonexistent.mp4`, name kind of implies it shouldn't be there but if that's the case the test should pass.

## Description
> [antek@Anteks-MBP]:[~/go/src/gocv.io/x/gocv] (getlayernames)$ make test
go test . ./contrib
--- FAIL: TestMSER (0.15s)
    features2d_test.go:176: Invalid KeyPoint array in MSER test: 232
OpenCV: Couldn't read video stream from file "nonexistent.mp4"
FAIL
FAIL	gocv.io/x/gocv	5.472s
ok  	gocv.io/x/gocv/contrib	0.727s
make: *** [test] Error 1

## Steps to Reproduce
1. cd $GOPATH/src/gocv.io/x/gocv
2. make test

## Your Environment
* Operating System and version:
OS X Mojave
* OpenCV version used:
4.0.0
* How did you install OpenCV?
brew install opencv
* GoCV version used:
dev branch @ 5a0e230
* Go version:
[antek@Anteks-MBP]:[~/go/src/gocv.io/x/gocv] (getlayernames)$ go version
go version go1.11.4 darwin/amd64
* Did you run the `env.sh` or `env.cmd` script before trying to `go run` or `go build`?
No | `env.sh` says its been deprecated

deadprogram:
Hello @Sauraus that is very interesting. We run the tests on both Linux and WIndows, but do not have a CI setup for macOS. Anyone else able to confirm tests failing on macOS?

Sauraus:
What is the intended test case here? Should the file exist or not?

jonathankarsh:
Hi @Sauraus and @deadprogram,
I'm also running on a mac. It seems that the failure occurs because of the KeyPoint array length in features2d_test.go (line 176). I also get `len(kp) == 232`. If I let this number through the if statement in TestMSER, this test passes.
The missing file `nonexistent.mp4` is OK. This is OpenCV complaining (so we get C++'s log message) but the related test passes. Would be nice to somehow make that message not appear, but not critical IMO.
Please LMK if you think adding `len(kp) == 232` to TestMSER makes sense, we can create a PR and close this issue off.

deadprogram:
Hello, all.

Is this still an issue with the latest versions of GoCV and OpenCV?

jonathankarsh:
[...]

PR #407, opened 2019-02-16
Add len(kp) == 232 to TestMSER, seems this is necessary for MacOS for…

jonathankarsh:
… some reason.

This addresses #389, but I really have no idea if this is a reasonable solution.


deadprogram:
The fact that the test now passes on macOS, Linux, and Windows seems like it indicates the test smell in the previous test. That said, it is probably better now, and certainly no worse. Merging, but seems like further research is probably needed.

PR #420, opened 2019-03-13
Added Bitwise Operations With Masks

sshibs:
The current bitwise operations don't have a parameter for a mask (I needed the mask parameter for a project), so I added new functions with this parameter - BitwiseAndWithMask, BitwiseNotWithMask, BitwiseOrWithMask, and BitwiseXorWithMask. Thanks!


deadprogram:
Thank you very much for this PR @sshibs it looks ideal. Now merging, thanks again.

PR #421, opened 2019-03-13
added BoxPoints function and BorderIsolated const

sshibs:
I added the BoxPoints function because it looks like it isn't implemented yet. Also, the BorderIsolated border type constant, was missing so I added that too.


deadprogram:
Thank you @sshibs another excellent looking PR. Now merging, we all appreciate your contribution.

PR #423, opened 2019-03-14
Addition of MOG2/KNN constructor with options

netbrain:
Hi guys! this is my first contribution to this project and I wanted to ensure that this code is correct in terms of the contributions guideline. So a little hand holding would be appreciated.

This change simply adds the the ability to change the default options that are set using the default MOG2 & KNN constructor.

Cheers!

### DIFF
--- a/video.cpp
+++ b/video.cpp
@@ -4,8 +4,16 @@ BackgroundSubtractorMOG2 BackgroundSubtractorMOG2_Create() {
     return new cv::Ptr<cv::BackgroundSubtractorMOG2>(cv::createBackgroundSubtractorMOG2());
 }
 
+BackgroundSubtractorMOG2 BackgroundSubtractorMOG2_CreateWithParams(int history, double varThreshold, bool detectShadows) {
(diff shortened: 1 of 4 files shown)


netbrain:
@deadprogram Hi, is it ok if i ask you for a code review?

deadprogram:
Hello, @netbrain thank you very much for the PR. I think you have the correct idea about what needs to be included. Two things I would suggest:

- Please always include the link to the OpenCV function that is being wrapped. For example https://github.com/hybridgroup/gocv/blob/0addb2293f9051055074fde2842f38c98ac7b36c/video.go#L20-L22 this makes it a lot easier for people to navigate from the GoDocs to the correct OpenCV docs.
- It probably would be best to have slightly more "atomic" PRs. Once you update your comments, I will be "squashing and committing" this PR, since there are multiple commits needed to achieve the desired code state. Hopefully that makes sense. No need to spilt this PR, I am just suggesting a bit of guidance for the next one. :)

netbrain:
Tank you @deadprogram, i thought as much myself. Ive gone and squashed it for you and also added the missing links to the actual function that my and the existing code calls.

Should be in tip top shape now i guess!

deadprogram:
Looks great, thanks for the changes. Now merging.

PR #426, opened 2019-03-17
initial cuda core and bgsegm implementation

netbrain:
Initial implementation of cuda core and cudabgsegm with MOG and MOG2 algorithms

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,6 +84,16 @@ build_nonfree:
 	$(MAKE) preinstall
 	cd -
 
+# Build OpenCV with cuda.
+build_cuda:
+	cd $(TMP_DIR)opencv/opencv-$(OPENCV_VERSION)
+	mkdir build
+	cd build
+	cmake -j $(shell nproc --all) -D CMAKE_BUILD_TYPE=RELEASE -D CMAKE_INSTALL_PREFIX=/usr/local -D OPENCV_EXTRA_MODULES_PATH=$(TMP_DIR)opencv/opencv_contrib-$(OPENCV_VERSION)/modules -D BUILD_DOCS=OFF -D BUILD_EXAMPLES=OFF -D BUILD_TESTS=OFF -D BUILD_PERF_TESTS=OFF -D BUILD_opencv_java=NO -D BUILD_opencv_python=NO -D BUILD_opencv_python2=NO -D BUILD_opencv_python3=NO -D WITH_JASPER=OFF -DOPENCV_GENERATE_PKGCONFIG=ON -DWITH_CUDA=ON -DENABLE_FAST_MATH=1 -DCUDA_FAST_MATH=1 -DWITH_CUBLAS=1 -DCUDA_TOOLKIT_ROOT_DIR=/usr/local/cuda/ -DBUILD_opencv_cudacodec=OFF ..
+	$(MAKE) -j $(shell nproc --all)
+	$(MAKE) preinstall
+	cd -
+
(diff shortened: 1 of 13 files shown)


netbrain:
@deadprogram Hi again! started on implementing cuda. Im on thin ground here, so please take a look if it looks right to you. Cheers.

netbrain:
I guess we need to have cuda compiled on the CI systems :(

go.mod
```diff
@@ -1 +1,3 @@
 module gocv.io/x/gocv
+
+go 1.12
```
netbrain: @deadprogram anyway to ignore the auto inclusion of the go version? Not sure what is adding this to this file, if it is go itself or my IDEA (goland).

netbrain:
@deadprogram looks like I need some input on how i can either exclude these cuda dependent files from compilation or add cuda support to travis and appveyor.

deadprogram:
The 2 options generally for this are either a build tag to exclude those files or else a sub-package.

In some cases, both, such as how the openvino sub-package is organized.

netbrain:
Tried a fast experiment where i created a new cuda package. However running the tests fails with `
cuda.h:5:10: fatal error: opencv2/opencv.hpp: No such file or directory
 #include <opencv2/opencv.hpp>
`
Im getting identical results when trying to run tests in the openvino package. I also tried with a custom environment but that seemed to create a whole lot of other errors.

Any idea what's going on? Ill look more into it later.

go.mod
netbrain: https://youtrack.jetbrains.com/issue/GO-7121

netbrain:
@deadprogram ping, just a friendly reminder :)

go.mod
deadprogram: So, you will remove that line manually? Seems like that might be your own option at the moment.

deadprogram:
Hi @netbrain this is an amazing PR. I have not had a chance to install the dependencies and test on my machine yet. 

As you have already identified, cuda would need to be installed on Travis and Appveyor in order to run the tests on the CI servers. However the needed access to physical hardware on those services does not exist. So that will not be happening anytime soon.

For those people who want to take the plunge, there should be a `cuda/README.md` file, describing what you need to do to install cuda on the various supported platforms. Also what version(s) of cuda are supported.

What do you think? Anyone else want to chime in here?

netbrain:
Thank you @deadprogram, for this first PR i would simply want to create the base platform for cuda features. As for the cuda/README file, sure i can add that, at least a minimal one. As you can see in the PR i have added a make target for installing opencv with cuda support, but i guess it should be possible to create a similar script that actually installs cuda itself aswell. But at the same time I really dont want to do too much in this PR, and instead add the missing pieces over time.

So to summarize, add a cuda/README file that guides the user to the cuda installation page (https://docs.nvidia.com/cuda/index.html) and document the cuda makefile target? 

Will that be sufficient for this PR?

deadprogram:
That sounds like a very reasonable plan to me.

netbrain:
@deadprogram  Added the readme. I think this should be good as a starting point for cuda contributions. I also refactored the names for some of the methods. They were prefixed with "Cuda" but i removed that as they now live in their own package.

netbrain:
@deadprogram so what is the plan with this? does it need more work?

deadprogram:
I think this is an incredible start for the community to be able to contribute to. Now merging, thank you for the great work!

PR #434, opened 2019-03-28
Add support for SimpleBlobDetectorParams

ryonsabouni:
Needed this functionality, and have started to use it. Open to any and all suggestions!

### DIFF
--- a/features2d.go
+++ b/features2d.go
@@ -356,6 +356,10 @@ func NewSimpleBlobDetector() SimpleBlobDetector {
 }
 
 // NewSimpleBlobDetectorWithParams returns a new SimpleBlobDetector with custom parameters
+//
+// For further details, please see:
+// https://docs.opencv.org/master/d0/d7a/classcv_1_1SimpleBlobDetector.html
+//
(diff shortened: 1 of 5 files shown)


features2d.go
```diff
@@ -350,13 +355,213 @@ func NewSimpleBlobDetector() SimpleBlobDetector {
 	return SimpleBlobDetector{p: unsafe.Pointer(C.SimpleBlobDetector_Create())}
 }
 
+// NewSimpleBlobDetectorWithParams returns a new SimpleBlobDetector with custom parameters
```
deadprogram: Can you please add the link to the OpenCV docs, same as that used by `NewSimpleBlobDetector()`? Thank you.

deadprogram:
This is really great contribution @ryonsabouni thank you for working on it. I made one small comment regarding the GoDocs. Otherwise looks good to merge.

features2d.go
ryonsabouni: Added

deadprogram:
Thank you very much for the PR and the update. Now merging.

PR #436, opened 2019-04-08
Use opencv4 on Windows

mattn:
I'm thinking gocv should use pkg-config always on Windows too. We can fix dependencies without modify code.


deadprogram:
I like this idea in theory. How do you install pkg-config for mingw-w64?

https://ci.appveyor.com/project/deadprogram/gocv/builds/23667840#L55

mattn:
If you don't mind pkg-config clone, `go get github.com/mattn/pkg-gonfig/pkg-config

mattn:
Test is passed on AppVeyor.

mattn:
Can you provide way to use pkg-config on Windows too?

mattn:
@deadprogram On Windows, we have to keep older version of OpenCV. (like opencv_dnn420) 
Could you provide way to use pkg-config?

mattn:
I'm thinking you've better to switch to use msys2/mingw64 for OpenCV4 on Windows since msys2/mingw64 provide pkg-config. If gocv remove the specifying cgo directive on Windows, you will be maintenance free for the versions of OpenCV4 on Windows.

PR #444, opened 2019-05-01
fix the command to install xquartz via brew/cask on MacOS

owulveryck:


deadprogram:
Thank you for the correction, @owulveryck 

I changed the branch to `dev` as mentioned here: https://github.com/hybridgroup/gocv/blob/master/CONTRIBUTING.md#how-to-use-our-github-repository

Now merging, thanks again.

PR #447, opened 2019-05-14
Fix possible resource leak

andriikushch:
Fix few places where 'defer' statements were called in a for loop. Calling 'defer' this way can cause a resource leak.


deadprogram:
Somehow missed this PR, but this is certainly useful. Thanks!

Now merging.

PR #464, opened 2019-06-04
added NewGpuMatFromMat which creates a GpuMat from a Mat

netbrain:


deadprogram:
Thank you for another GPU oriented contribution. Now merging...

Issue #395, opened 2019-02-04
Implement blobFromImages


WendelHime:
I've seen documentation and code and didn't find the function blobFromImages, it shouldn't be on ROADMAP.md or someone is already implementing?

deadprogram:
Hi @WendelHime you are correct, it needs to be added to roadmap and also implemented.

sukhoi:
The signature in V2 and V3 is different from V4. Any suggestions/clarifications/workarounds for that?
@WendelHime : In case you haven't started on this one, I'd like to pick this up. I started without checking if someone was already working :(. Please let me know.
Thanks. -H

WendelHime:
Hi @sukhoi , I need to learn more about the communication between languages and OpenCV C++, so I didn't try.

deadprogram:
Hello @sukhoi I suggest target only OpenCV 4 since the rest of GoCV is currently doing that now.

To use OpenCV 3.4 a person will need to use an older version of GoCV now anyhow.

sukhoi:
@deadprogram : Would it be wise to put placeholders/signatures for the methods that needed to be implemented so that who ever picks up the code does not worry about the version, syntax, source reference, etc.

PR #467, opened 2019-06-12
add BlobFromImages

WendelHime:
Hi guys, I'm trying to implement the #395 , someone can please review and make suggestions?


WendelHime:
@deadprogram should I refactor blobFromImage to follow the [opencv 4.1 specification](https://docs.opencv.org/master/d6/d0f/group__dnn.html#ga98113a886b1d1fe0b38a8eef39ffaaa0)?

deadprogram:
There are 2 signatures as that function is overloaded and we already have implemented one of them. There is has been no change in that version, so the answer would be "no, that is not needed".

deadprogram:
Also, thank you for the contribution! Now squashing/merging.

Thanks again!

PR #478, opened 2019-07-04
format: better format

xpzouying:
use `fmt.Printf` to print %v


deadprogram:
Thank you for the correction, @xpzouying 

I changed the branch to `dev` as mentioned here:
https://github.com/hybridgroup/gocv/blob/master/CONTRIBUTING.md#how-to-use-our-github-repository

Now merging. Thanks again!

PR #483, opened 2019-07-12
RFC: Add KMeans function

acharyab15:
Added the KMeans function.

However, I also realized that I wanted to call the function on a `[]image.Point` to cluster a bunch of contour points and not just a `Mat` type; so my solution was to create a `KMeansPoints` function.

Please let me know if there is a better way to do it and that function isn't required. If it is required, please let me know if you would like to rename it to something better.

### DIFF
--- a/core.cpp
+++ b/core.cpp
@@ -485,6 +485,21 @@ double Mat_Invert(Mat src, Mat dst, int flags) {
     return ret;
 }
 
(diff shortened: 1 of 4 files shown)


deadprogram:
Seems like a reasonable approach. Thank you for the contribution @acharyab15 now merging.

PR #487, opened 2019-07-18
added MeanWithMask function for Mats

sshibs:
The Mean() function does not have the additional parameter for a mask, so I added a separate function MeanWithMask(). Thanks!

### DIFF
--- a/core.go
+++ b/core.go
@@ -374,9 +374,20 @@ func (m *Mat) ConvertFp16() Mat {
 }
 
 // Mean calculates the mean value M of array elements, independently for each channel, and return it as Scalar
-// TODO pass second paramter with mask
+// For further details, please see:
+// https://docs.opencv.org/master/d2/de8/group__core__array.html#ga191389f8a0e58180bb13a727782cd461
+//
 func (m *Mat) Mean() Scalar {
 	s := C.Mat_Mean(m.p)
+	return NewScalar(float64(s.val1), float64(s.val2), float64(s.val3), float64(s.val4))
+}
+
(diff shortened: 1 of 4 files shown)


core.go
```diff
@@ -374,12 +374,23 @@ func (m *Mat) ConvertFp16() Mat {
...
 	s := C.Mat_Mean(m.p)
 	return NewScalar(float64(s.val1), float64(s.val2), float64(s.val3), float64(s.val4))
 }
 
+// MeanWithMask calculates the mean value M of array elements,independently for each channel,
+// and returns it as Scalar vector while applying the mask.
+// https://docs.opencv.org/master/d2/de8/group__core__array.html#ga191389f8a0e58180bb13a727782cd461
+//
+func (m *Mat) MeanWithMask(mask Mat) Scalar {
+	s := C.Mat_Mean(m.p, mat.p)
```
acharyab15: Shouldn't this call Mat_MeanWithMask? And what is the variable `mat`? Shouldn't that be `mask`? I don't understand why the build is not failing though.

sshibs: Oh wow, thank you for catching those. I think I must have been half asleep when I did this yesterday. Sorry about that! Fixing it now.

For some reason the build was passing even though Mat_Mean only took 1 argument and I was passing 2 arguments to it. when I changed it to Mat_MeanWithMask it did fail because of the mis-named mask argument. Kind of weird... I guess must be something about go's C library?

acharyab15: No worries! Yeah, it was really strange how it didn't catch that!

deadprogram:
Thank you for the PR, @sshibs now going to squash/merge.

Thanks again!

PR #498, opened 2019-08-20
Example: face detection from image url by using IMDecode

bookun:
I Added an example.

There are not any example using IMDecode.

### DIFF
--- a/cmd/facedetect-from-url/main.go
+++ b/cmd/facedetect-from-url/main.go
@@ -1,0 +1,79 @@
+// What it does:
+//
+// This example uses the CascadeClassifier class to detect faces from url,
+// and draw a rectangle around each of them, before displaying them within a Window.
+//
+// How to run:
+//
+// facedetect-from-url [image URL] [classifier XML file] [image file]
+//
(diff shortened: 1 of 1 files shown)


cmd/facedetect-from-url/main.go
```diff
@@ -0,0 +1,79 @@
+// What it does:
+//
+// This example uses the CascadeClassifier class to detect faces from url,
+// and draw a rectangle around each of them, before displaying them within a Window.
+//
+// How to run:
+//
+// facedetect-from-url [image URL] [classifier XML file] [image file]
+//
+// 		go run ./cmd/facedetect/main.go https://raw.githubusercontent.com/hybridgroup/gocv/master/images/face.jpg data/haarcascade_frontalface_default.xml output.jpg
```
deadprogram: This line needs to be updated to correctly point to the location of this new example, and to use the correct syntax.

deadprogram:
Hi @bookun thank you for the contribution.

This PR should be made against the `dev` branch as mentioned in our guidelines here:
https://github.com/hybridgroup/gocv/blob/master/CONTRIBUTING.md#how-to-use-our-github-repository

Thanks!

PR #499, opened 2019-08-27
Example: face detection from image url by using IMDecode

bookun:
I deleted the branch of #498 by mistake....

I fixed review point from @deadprogram .
Thanks for the review.


deadprogram:
Thanks for the correction/addition @bookun now squash/merging.

PR #522, opened 2019-10-16
Added InvertAffineTransform

dlion:


deadprogram:
Thanks for another useful contribution @dlion now merging.

PR #523, opened 2019-10-18
Fix webp image decode error

chenttuuvv:
when decode webp image like this:

```
package main
import "gocv.io/x/gocv"
import "io/ioutil"

func main() {

        fileData,err := ioutil.ReadFile("test.webp")
        if err != nil {
                panic(err)
        }

        gocv.IMDecode(fileData, gocv.IMReadUnchanged)

}

```

We met error:
imdecode_(''): can't read data: OpenCV(4.1.0) /data/opencv/opencv-4.1.0/modules/imgcodecs/src/grfmt_webp.cpp:164: error: (-215:Assertion failed) data.type() == CV_8UC1 in function 'readData'

we resolve this problem by changing vector template type from char to uchar, align with function Image_IMEncode_WithParams in imgcodecs.cpp

reference:
https://answers.opencv.org/question/210682/cv2-imdecode-webp-format-fails/


deadprogram:
That is one subtle fix @chenttuuvv thank you very much for finding that issue.

What do you think about also adding a test for this? You could add an image in `webp` format. It is not strictly necessary but would prove that it works as expected.

chenttuuvv:
@deadprogram Here is the sample image we used to test imdecode function
https://i0.hdslb.com/bfs/album/293ee603a4d5f1581655b9715acd9e82bcdd8deb.jpg@2000w_1e.webp

deadprogram:
@chenttuuvv can you please add the test to this PR? You can either rebase or I can squash the commits for you once it is ready to merge. Thank you.

chenttuuvv:
@deadprogram please check it out

deadprogram:
That looks great @chenttuuvv thank you for adding the test. Now squash/merging, thanks again for your contribution!

PR #525, opened 2019-10-19
Added ClipLine

dlion:


deadprogram:
Thanks for the PR @dlion looks good now merging.

PR #343, opened 2018-10-19
core: add ScaleAdd() method to Mat

morfeo8marc:
Add ScaleAdd method for enhancement request #341.

PR #528, opened 2019-10-21
Update ROADMAP.md by removing scaleAdd

toffentoffen:
Update ROARDMAP.md as scaleAdd has been alrady implemented in #343.


deadprogram:
The pattern I have been using is to mark features with an `X` once it lands in the `dev` branch, then to remove after it makes it into a release aka `master`. That way if someone looks at the `dev` branch they can figure out what is available but not yet released. 

This info should probably be added to the CONTRIBUTING.md, my bad.

toffentoffen:
makes total sense @deadprogram. The problem is that this function is already in master, therefore I just removed the line. We didn't update properly the file last year when pr #343

toffentoffen:
Done @deadprogram can be merged :smile:

deadprogram:
Thank you! Now merging.

Issue #529, opened 2019-10-21
Add setIdentity operation to Arrays


toffentoffen:
Hi,

I will start working on adding the setIdentity method.

deadprogram:
Marked for next release, I will close this issue once it is available. Thanks!

deadprogram:
This was part of release v0.22.0 so now closing. Thank you!

PR #530, opened 2019-10-21
core: add setIdentity() method to Mat

toffentoffen:
Add SetIdentity method for enhancement request #529


deadprogram:
Thanks for another useful contribution @toffentoffen we appreciate it! Now merging.

Issue #537, opened 2019-10-31
FillPoly not getting expected result


huichen:
## Description
FillPoly is not getting expected result (a filled rectangle). What I'm getting after running the code is this image:

[test.png](https://postimg.cc/0zjFzrNC)

## Steps to Reproduce

```
package main

import (
        "image"
        "image/color"
        "log"

        "gocv.io/x/gocv"
)

func main() {
        img := gocv.NewMatWithSize(100, 100, gocv.MatTypeCV8UC3)
        defer img.Close()

        poly := [][]image.Point{
                {
                        image.Pt(10, 10),
                        image.Pt(10, 20),
                        image.Pt(20, 20),
                        image.Pt(20, 10),
                },
        }
        log.Printf("%+v", poly)
        gocv.FillPoly(&img, poly, color.RGBA{0, 255, 0, 0})
        gocv.IMWrite("test.png", img)

        log.Printf("done")
}

```

## Your Environment

* Operating System and version: Ubuntu 18.04
* OpenCV version used: 4.1.1
* How did you install OpenCV? from source
* GoCV version used: 0.21.0
* Go version:1.12.5
* Did you run the `env.sh` or `env.cmd` script before trying to `go run` or `go build`? yes

deadprogram:
I get something else yet again:

![test](https://user-images.githubusercontent.com/5520/68325861-9451bd00-00ca-11ea-96e3-8c9ef5bfd4df.png)

huichen:
https://github.com/hybridgroup/gocv/blob/master/imgproc.go#L1233-L1250

I believe defer C.free at line 1235 is called before leaving the loop, and variable p has undetermined value inside C.FillPoly at line 1265, thus we are getting different and both unexpected result.

Any reason wrapping it in a func(){} ?

Similar problem for DrawContours function.

deadprogram:
@huichen you are absolutely correct. That memory cannot be freed until after the call to `C.FillPoly()` returns.

The only reason to wrap that in a func would be to use `defer` to release the memory, but as you have already established, that is incorrect.

I'd gladly accept a PR that corrects this, and also the same for `DrawContours()`.

huichen:
Thanks for the clarification.  Just submitted a PR. Please take a look.

deadprogram:
This issue should now have been addressed by #543 thank you @huichen for the fix. Now closing this issue, please reopen if needed.

PR #543, opened 2019-11-14
In FillPoly and DrawContours functions, remove func() wrap to avoid memory freed before calling opencv functions.

huichen:
to solve issue

https://github.com/hybridgroup/gocv/issues/537


ibadami:
Any update on this PR?

deadprogram:
`defer` can be tricky. Seems like a correct change, now merging. Thank you @huichen for the update.

ibadami:
@deadprogram Thanks for the awesome speed of reply!

PR #545, opened 2019-11-20
Tolerate imprecise result in SolvePoly

siddhesh:
The SolvePoly test TestSolvePoly generates a result that is slightly
imprecise on aarch64 (~2.75e-62) as opposed to a precise result on
x86_64.  This change adjusts the test to tolerate this imprecision.


deadprogram:
Thanks for the PR @siddhesh now merging.

PR #552, opened 2019-11-24
add String method

mh-cbon:
it is not automated as i wanted initially, but go stringer is not as useful as i thought, and, i was not feeling to rewrite stringer (yet). anyway, it is done.


deadprogram:
Hi @mh-cbon thanks for the PR. Took me a minute to realize what this was doing. That was a lot of work!

Only request I have, I am pretty sure that the file` contrib/data.yaml` is generated by the contrib tests, and probably should not be committed to the repo and also should probably added to the `.gitignore` files. 

What do you think?

mh-cbon:
all done, i also updated with upstream. thank you for the feedback and merging.

deadprogram:
Thank you! Now squash/merging.

PR #568, opened 2019-12-12
Add openvino async

Danile71:


deadprogram:
Hi @Danile71 thank you for the PR. Just wondering if the files would be better named `dnn_async_openvino.go` etc. since this is functionality that only works with openvino.

Since it is part of the dnn module, I can see why you would not want to put into the `openvino` subpackage.

Anyhow, what do you think?

Danile71:
Hi, @deadprogram,  I tried to place *async to another packer (openvino).
But `func (net *Net) ForwardAsync(outputName string) AsyncArray` must be in package gocv (net *gocv.Net not working)

Danile71:
@deadprogram, you are right)

deadprogram:
Shoudn't `asyncarray.cpp` be named `dnn_asyncarray_openvino.cpp`
`asyncarray.go` be named `dnn_asyncarray_openvino.go`, and so on for the rest of the newly added files?

Danile71:
`opencv2/core/async.hpp`
It is "core" in official sources opencv, not dnn

deadprogram:
OK, fair enough. Now merging, thank you!

Issue #191, opened 2018-05-12
[Feature] Make gocv.VideoCapture thread-safe


denismakogon:
So, the idea to make `gocv.VideoCapture` safe while accessing one from goroutines.
Here's the use case: as developer, i want to parse a video in frames (1 chunk equals to video's FPS) concurrently, the most efficient way to make this happen to  use `VideoCapture.Set` and `VideoCapture.Read`.
This is a sample code to do a frame reading (not thread-safe):
```go
func readFramesFromPosition(video *gocv.VideoCapture, startFrameIndex, framesToRead int64) ([]*gocv.Mat, error) {
	log := logrus.New()
	var mats []*gocv.Mat
	tmpMat := gocv.NewMat()

	video.Set(gocv.VideoCapturePosFrames, float64(startFrameIndex))
	defer video.Set(gocv.VideoCapturePosFrames, float64(0))
	log.Infof("reading frames for range: [%d:%d]", startFrameIndex, startFrameIndex+framesToRead)

	for i := int64(0); i < framesToRead; i++ {
		success := video.Read(&tmpMat)
		if !success {
			return nil, fmt.Errorf("unable to read frame on position: '%d' "+
				"from a video file", startFrameIndex+i+1)
		}
		mats = append(mats, &tmpMat)
	}
	return mats, nil
}
```
So, all i want is to start a bunch of these functions in parallel to speed up processing time because it make take hell of a time to process all frames from, let's say 1.4Gb, movie file.

My suggestion is:
 - make an acess to `Set` and `Read` mutex-protected, similar to `gocv.VideoWriter`

breaf solution:
```go
import (
	"sync"

	"gocv.io/x/gocv"
)

type SyncVideoCapture struct {
	mu *sync.RWMutex
	v gocv.VideoCapture
}

func (sv *SyncVideoCapture) Set(prop gocv.VideoCaptureProperties, param float64) {
	sv.mu.Lock()
	sv.v.Set(prop, param)
	sv.mu.Unlock()
}

func (sv *SyncVideoCapture) Get(prop gocv.VideoCaptureProperties) float64 {
	sv.mu.Lock()
	defer sv.mu.Unlock()
	return sv.v.Get(prop)
}

func (sv *SyncVideoCapture) Read(m *gocv.Mat) bool {
	sv.mu.Lock()
	defer sv.mu.Unlock()
	return sv.v.Read(m)
}
```

denismakogon:
[...]

PR #570, opened 2019-12-12
Fix crash when two or more windows

Danile71:


Danile71:
I founded bug:
When I start recieving video from 2 cameras and parallel drawing to 2 windows - I have crash.
With sync no bugs

deadprogram:
Makes sense given the way that OpenCV does all of the UI stuff thru a single thread. I think I tried doing something like this once, not sure why I did not continue?

What OS do you use @Danile71 for testing?

deadprogram:
This was one of the issues where thread safety was discussed: https://github.com/hybridgroup/gocv/issues/191

Danile71:
I use ubuntu 18

PR #571, opened 2019-12-12
Add new functions

Danile71:


deadprogram:
Hi @Danile71 thank you for working on this contribution.

2 things that need to be corrected. The first, is that you need to run go mt on the code: https://travis-ci.org/hybridgroup/gocv/builds/627156392#L862

The second is that this new code needs some test coverage, please.

Thank you very much.

PR #572, opened 2019-12-12
Add support QR codes

Danile71:


deadprogram:
Thanks @Danile71 I had not even noticed that this had been added to OpenCV to put on the roadmap!

Now merging, thank you for the contribution.

PR #576, opened 2019-12-18
imgcodecs: optimize copy of data used for IMDecode method

deadprogram:
This PR optimizes the copying of data used for IMDecode method, to improve in-memory only performance.


deadprogram:
Any opinions on this PR?

PR #577, opened 2019-12-18
dnn: respect the bit depth of the input image to set the expected output

deadprogram:
This PR respects the bit depth of the input image to set the expected output when converting an image to a blob. Needed to allow for proper DNN handling of greyscale images.

### DIFF
--- a/dnn.cpp
+++ b/dnn.cpp
@@ -97,10 +97,17 @@ void Net_GetUnconnectedOutLayers(Net net, IntVector* res) {
 Mat Net_BlobFromImage(Mat image, double scalefactor, Size size, Scalar mean, bool swapRB,
                       bool crop) {
     cv::Size sz(size.width, size.height);
-    cv::Scalar cm = cv::Scalar(mean.val1, mean.val2, mean.val3, mean.val4);
 
-    // TODO: handle different version signatures of this function v2 vs v3.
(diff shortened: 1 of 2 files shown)


deadprogram:
This line of code results in a compiler warning because of the NULL parameter:

```
return new cv::Mat(cv::dnn::blobFromImage(*image, scalefactor, sz, NULL, swapRB, crop, ddepth));
```

I was not able to figure out another way to pass something other than `Scalar()` which results in an error when you try to create a blob from an image with color depth of `CV_8U` e.g. grayscale image.

Perhaps someone else has an idea on how to do this, or perhaps this is in fact the correct way.

eh-steve:
> I was not able to figure out another way to pass something other than `Scalar()` which results in an error when you try to create a blob from an image with color depth of `CV_8U` e.g. grayscale image.
> 
> Perhaps someone else has an idea on how to do this, or perhaps this is in fact the correct way.

What error do you get when you use `Scalar()` for `CV_8U` `Mat`s? I tried it locally and it worked, and seems to align with this logic: https://github.com/opencv/opencv/blob/master/modules/dnn/src/dnn.cpp#L324

hbmask:
```
//        return new cv::Mat(cv::dnn::blobFromImage(*image, scalefactor, sz, NULL, swapRB, crop, ddepth));
        return new cv::Mat(cv::dnn::blobFromImage(*image, scalefactor, sz, cv::Scalar(), swapRB, crop, ddepth));
```

Issue #486, opened 2019-07-18
Memory Leak in Mat.Size()


sshibs:
I am using gocv to process a large number of images and I noticed I had a bit of a memory leak. I narrowed it down to the Mat.Size() call. It could be you already know about this, and I don't know what the solution would be, but I thought I would post it just incase.

## Description

Looking at the C++ code for Mat_Size it looks like the issue is because "new" is called to make the ids array, but it is never deleted because ids is returned to the go code.

```
void Mat_Size(Mat m, IntVector* res) {
    cv::MatSize ms(m->size);
    int* ids = new int[ms.dims()];

    for (size_t i = 0; i < ms.dims(); ++i) {
        ids[i] = ms[i];
    }

    res->length = ms.dims();
    res->val = ids;
    return;
}
```

## Steps to Reproduce

Here is a simple program to reproduce the issue:
```
import (
	"gocv.io/x/gocv"
)

func main() {
	img := gocv.NewMatWithSizeFromScalar(gocv.Scalar{0, 0, 0, 0}, 500, 500, gocv.MatTypeCV8UC3)
	defer img.Close()
	for i := 0; i < 100000000; i++ {
		img.Size()
	}
}
```

## Your Environment

* Operating System and version: Ubuntu 18.04
* OpenCV version used: 4.1
* How did you install OpenCV? installation guide
* GoCV version used: 0.20.0
* Go version: 1.12.1
* Did you run the `env.sh` or `env.cmd` script before trying to `go run` or `go build`? No

deadprogram:
This issue was corrected by #580 so now closing. Thank you!

PR #580, opened 2019-12-24
 Fix memory leak in Mat.Size() and Mat.Split()

### LINKED ISSUE #394: memory leak when call gocv.split()
## Description

example:
Assum that img is a 2 channel Mat

for {
    imgarr := gocv.splite(img)
    imgarr[0].Close()
    imgarr[1].Close()
}
Kind of above codes cause memory leak. I am sure that the Mat objects in imgarr are closed, but the memory usage is still growing up. Why?

## Steps to Reproduce

1. https://stackoverflow.com/questions/54456414/garbage-collection-is-not-working-with-gocv-mat-in-golang/54457054#54457054
2.
3.
4.

## Your Environment

* Operating System and version:win10
* OpenCV version used:4.0
* How did you install OpenCV?as official site instructions
* GoCV version used:4.0
* Go version:1.11.5
* Did you run the `env.sh` or `env.cmd` script before trying to `go run` or `go build`?no

xxb-at-julichina:
This patch fix #394 and #486.

### DIFF
--- a/core.cpp
+++ b/core.cpp
@@ -761,3 +761,6 @@ Mat Mat_colRange(Mat m,int startrow,int endrow) {
     return new cv::Mat(m->colRange(startrow,endrow));
 }
 
+void IntVector_Close(struct IntVector ivec) {
+    delete[] ivec.val;
+}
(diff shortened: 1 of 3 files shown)


deadprogram:
Thank you for the contribution @xxb-at-julichina now squashing/merging. Thanks again!

PR #587, opened 2019-12-30
Dev fix warn

Danile71:


deadprogram:
So the underlying problem is this line of code in OpenCV:
https://github.com/opencv/opencv/blob/master/modules/dnn/src/dnn.cpp#L286

Basically, in the case that the image is grayscale aka `CV_8U` then you cannot pass a `Scalar` to the mean value. However, if a NULL is pass, you get the compiler warning.

Perhaps there is some other value that can be passed, such as one of the base classes for `Scalar`? That was as far as I got.

Also, just to mention that the compiler warning is only on the CGo compile, not on execution of the resulting code.

deadprogram:
I have already addressed this issue in the `dev` branch, so now closing. Thank you!

PR #588, opened 2020-01-02
Added raspberry pi zero support to the makefile

scott97:
The raspberry pi zero uses a different instruction set to the raspberry pi 2 & 3. I have added to the makefile 2 new commands, **build_raspi_zero** and **install_raspi_zero**, for installing on the rpi zero.


deadprogram:
Thanks for this contribution @scott97 now merging.

Issue #105, opened 2018-02-26
cv::Point2f data structure in Go?


willdurand:
Hi,

`gocv` relies on the Go [`image.Point`](https://golang.org/pkg/image/#Point) (X and Y are integers) but it seems opencv uses `cv::Point2f` time to time and we do not have a Go data structure for it. Should we add it or is there something I've overlooked?

Cheers,
William

berak:
@willdurand , if you want my 2ct, -- [do it  !!](https://github.com/opencv/opencv/search?utf8=%E2%9C%93&q=Point2f&type=) 

from there on, it would probably be a good idea, to have Mat setters / getters for this, so we don't need an additional C structure for `vector<Point2f>` or `vector<vector<Point2f>>`

again, opencv's own python/java wrappers treat a `vector<Point2f>` simply as a `Mat(h,w,CV_32FC2)`, and a `vector<vector<Point2f>>` would translate to a `struct Mats` 
(very much simplifying the internal wrapping, and leaving the problem, how to access items to userland)

willdurand:
This is an interesting approach, I like it.

willdurand:
@deadprogram could you assign me to this? I'd like to give it a shot

deadprogram:
Done... go get 'em, tiger! :)

deadprogram:
Just wonder if the recent #139 sufficiently addresses this, or if more needs to be done.

@willdurand @berak what do you think?

deadprogram:
Pinging to update this issue... what is the status on it?

deadprogram:
This was actually implemented long ago. Oops! Now closing.

PR #589, opened 2020-01-02
Add Go Point2f type and update GetPerspectiveTransform()

eh-steve:
- Adds a Go `Point2f` struct with float32 X/Y fields
- Adds a C `Points2f` struct wrapper (with `Contour2f` alias)
- Updates `GetPerspectiveTransform()` to use new `[]Point2f` type instead of `[]image.Point` - this is an API-breaking change, but the original type signature did not match the C++ type signature, since the C++ expects a `cv::Point2f`.

Possibly addresses https://github.com/hybridgroup/gocv/issues/105


deadprogram:
Hi @eh-steve thanks for the PR and sorry about the delay in getting you some feedback.

Whenever possible, I have tried to use the Go std lib for interfaces to OpenCV, then converting to the needed OpenCV type. This has made it generally easier when integrating with some other non-OpenCV but image related Go code.

This is why the current code converts to `std::vector<cv::Point2f>` in the current code.

Is there something specific that you need to do that this is preventing you from doing?

If so, then it might be better to add a separate function such as `GetPerspectiveTransform2f()` to avoid breaking the current API.

What do you think?

eh-steve:
Yeah, I appreciate that Go’s `image.Point` type is handy and works well for most things which use integer pixels, but this function in particular often need floats rather than ints as inputs (e.g. when transforming between pixel space and a Cartesian or GPS coordinate system).

I’m happy to add a second function which uses floats to avoid breaking the API, but I think the new Point2f types will have to stay, since Go’s `image` package doesn’t have a float point type.

Do you want me to add a second function for float points?

deadprogram:
> Do you want me to add a second function for float points?

Yes, please.

All good points (ducks) sounds like a plan.

imgproc.go
```diff
@@ -1602,14 +1602,25 @@ func ApplyCustomColorMap(src Mat, dst *Mat, customColormap Mat) {
...
-// corresponding 4 point pairs.
+// corresponding 4 point pairs as image.Point.
 //
 // For further details, please see:
 // https://docs.opencv.org/master/da/d54/group__imgproc__transform.html#ga8c1ae0e3589a9d77fffc962c49b22043
 func GetPerspectiveTransform(src, dst []image.Point) Mat {
 	srcPoints := toCPoints(src)
 	dstPoints := toCPoints(dst)
-	return newMat(C.GetPerspectiveTransform(srcPoints, dstPoints))
+	return Mat{p: C.GetPerspectiveTransform(srcPoints, dstPoints)}
```
deadprogram: Any reason to not use `newMat()` here to be able to use profiling?

imgproc_test.go
```diff
@@ -1385,6 +1385,31 @@ func TestGetPerspectiveTransform(t *testing.T) {
 	}
 }
 
+func TestGetPerspectiveTransform2f(t *testing.T) {
+	src := []Point2f{
+		{0, 0},
+		{10, 5},
```
deadprogram: Should these be various floating point values here so as to smoke test with more "realistic" data?

imgproc.go
eh-steve: Good catch, I missed that

imgproc_test.go
eh-steve: I've updated the test, although they definitely are interpreted as `float32`s in the struct literal, even if they don't have a decimal point

deadprogram:
Thanks for all of the work on this PR @eh-steve now merging.

PR #602, opened 2020-02-04
Refactored SelectRoi methods

netbrain:


deadprogram:
Very late to the party on this PR, but seems like a very good idea @netbrain thank you for another cool contribution. Sorry I did not merge already, but now rectifying that oversight by merging now.

Thanks again!

PR #607, opened 2020-02-07
Remove breaking case statement

Danile71:


deadprogram:
Makes sense. Merging.

PR #610, opened 2020-02-07
Add cuda backend/target

Danile71:


deadprogram:
Thanks for this PR @Danile71 now merging.

PR #612, opened 2020-02-07
Fix build cuda and openvino

Danile71:


deadprogram:
Took me a moment to realize that this PR is literally to handle the case that you wish to build something for BOTH cuda and openvino.

Now merging, thanks for the contribution @Danile71

PR #614, opened 2020-02-10
update Makefile

Danile71:
@deadprogram, please check it, what about it?


deadprogram:
This is a lot of different things wrapped up into one PR. However, we do need all of them, so I am going to merge anyhow.

Thanks for the contribution(s) @Danile71 more excellent work!

PR #620, opened 2020-02-23
Threshold() should return a value

mdtsandman:
I'm new to both opencv and gocv.

The [OpenCV Threshold(...) function()](https://docs.opencv.org/3.3.0/d7/d1b/group__imgproc__misc.html#gae8a4a146d1ca78c626a53577199e9c57) returns a threshold value for use in later computations (eg. Canny edge detection). The [analogous GoCV function](https://pkg.go.dev/gocv.io/x/gocv?tab=doc#Threshold) operates on the destination Mat, but returns no value.

Hopefully this small change won't break things? I've used the float32 type for the returned value, but if it should be float64 please feel free to change.

Thanks for all your work on this fantastic set of bindings!

Mike Traynor.

### DIFF
--- a/imgproc.cpp
+++ b/imgproc.cpp
@@ -382,8 +382,8 @@ void Integral(Mat src, Mat sum, Mat sqsum, Mat tilted) {
     cv::integral(*src, *sum, *sqsum, *tilted);
 }
 
-void Threshold(Mat src, Mat dst, double thresh, double maxvalue, int typ) {
-    cv::threshold(*src, *dst, thresh, maxvalue, typ);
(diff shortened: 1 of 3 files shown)


deadprogram:
Hi @mdtsandman thank you for your contribution.

According to https://travis-ci.org/hybridgroup/gocv/builds/654170795#L861 the CI tests failed due to code formatting. Could you please run `go fmt` on your changes and update this PR?

Thanks!

mdtsandman:
Go fmt run as requested. Errors corrected.

deadprogram:
Thanks again for the PR, now squashing/merging.

PR #625, opened 2020-02-26
Added Polylines feature

golubaca:


deadprogram:
Thank you very much for the contribution @golubaca now merging.

PR #626, opened 2020-02-27
Add PhaseCorrelate binding

eh-steve:


golubaca:
Inside imgproc.go, you are receiving copies of Mat objects, not the pointers, which is not closed after function does it's things. Tests are failing because of that, which will lead to memory leaks if not fixed. Try receiving pointers instead of copies. And please, merge to dev branch instead of master branch.

eh-steve:
Thanks for looking at the PR!

I thought Go pointers were to be used to indicate to Go developers that the `Mat` might be mutated, whereas non-pointers could be assumed to be treated as read only (even though both structs contain C pointers to the same underlying mutable `cv::Mat`). In this case, all 3 inputs are read only, and so don't need to be pointers.

The tests were failing because I forgot to close the `Mat` used for the `window` parameter, which is now fixed

deadprogram:
This PR now has merge conflicts that needs to be resolved against the `dev` branch, please.

deadprogram:
Tests failing due to call the "wrong" version of `GetPerspectiveTransform()`.

deadprogram:
Thanks for another PR @eh-steve now squash/merging.

PR #631, opened 2020-03-05
Fix DNN warning

eh-steve:
Use `cv::Scalar()` instead of NULL to fix warning in `dnn.cpp`


deadprogram:
I addressed this thru other means, which has now been merged into the `dev` branch. Hence, closing this PR.

PR #642, opened 2020-03-20
imgproc: do not free contours data until after we have drawn them

deadprogram:
This PR is to ensure we do not free contours data until after we have drawn the needed contours.


deadprogram:
Now merging.

PR #655, opened 2020-04-09
Wrong documentation [cuda.go]

omriLugasi:
The path 
`https://github.com/opencv/opencv` does not exist
The relevant path should be?
`https://github.com/opencv/opencv`


deadprogram:
Hi @omriLugasi you are quite correct. Thanks for the contribution. Now merging.

PR #668, opened 2020-04-26
Feature/defer net close

frankegoesdown:
some refactor with defer calls


Apokalypser:
looks reasonable to me?

Issue #658, opened 2020-04-12
gocv.BlobFromImage is not work well.


knight2995:
## Description
I am trying to test dnn-detection [Code](https://github.com/hybridgroup/gocv/blob/master/cmd/dnn-detection/main.go)
. I am receiving an error message 
`gocv.io/x/gocv._Cfunc_Net_BlobFromImage(0x470f010, 0x3f80101010101010,...):`

I think the scalefactor parameter of the BlobFromImage seems to be wrong.
Only in 1.0 does not the error occur.

## Steps to Reproduce
1. I use TensorFlow Model and pbtxt config file.

## Your Environment

* Operating System and version: Windows 10
* OpenCV version used: 4.2.0
* How did you install OpenCV? No
* GoCV version used: 0.22.0
* Go version: 1.13.5
* Did you run the `env.sh` or `env.cmd` script before trying to `go run` or `go build`?

deadprogram:
I think that the underlying API of OpenCV changed or else the model itself did, as I ran into a similar problem with scaling the input blob to a DNN recently. I will look into it further.

knight2995:
Hi, I found this code in [opencv/dnn/dnn.cpp](https://github.com/opencv/opencv/blob/master/modules/dnn/src/dnn.cpp)

```
if (ddepth == CV_8U)  
    {
        CV_CheckEQ(scalefactor, 1.0, "Scaling is not supported for CV_8U blob depth");
        CV_Assert(mean_ == Scalar() && "Mean subtraction is not supported for CV_8U blob depth");
    }
```
I think this is the reason for the error. 

so, I try to change CV_8U to CV_32F, this code is worked fine.

```
// fix code
img.ConvertTo(&img, gocv.MatTypeCV32F)

blob := gocv.BlobFromImage(img, ratio, image.Pt(300, 300), mean, swapRGB, false)
```

craigske:
Seeing this issue as well. Even converting doesn't work for me.
img.ConvertTo(&imgBlob, gocv.MatTypeCV32F)
		blob := gocv.BlobFromImage(imgBlob, 1.0, image.Pt(imgBlob.Rows(), imgBlob.Cols()), gocv.NewScalar(104, 117, 123, 0), false, false)

gives 
[...]

PR #670, opened 2020-05-02
I change cmd code.

knight2995:
I write [Issue](https://github.com/hybridgroup/gocv/issues/658#issue-598454112).
I fix the error by creating an image with a data type of CV_32F from existing Image before creating Blob. 
the changed code works fine, but I wonder if this is correct.


EricaseJang:
I think you should close the cloned image Mat.

Issue #674, opened 2020-05-19
Support detectAndDecodeMulti for QRCodeDetector


OmriTreidel:
## Description
OpenCV  QRCodeDetector supports detection  and encoding of multiple QR codes. The goCV wrapper doesn't have a wrapper for it. It would be great to have it.

OmriTreidel:
@deadprogram I've created a PR for this issue  a few months ago. Any chance of someone looking at it?

OmriTreidel:
bump

deadprogram:
This has now been released so closing. Thanks!

PR #680, opened 2020-05-21
wrapper for detectAndDecodeMulti

OmriTreidel:
This PR adds a wrapper around `cv::QRCodeDetector::detectAndDecodeMulti` and `cv::QRCodeDetector::detectMulti` requested in issue #674 

In the process a conversion between CStrings to Go strings was needed. Interestingly the code in dnn.GetLayerNames didn't work since the strings were not contiguous in memory so a `toGoStrings` method was implemented and used in GetLayerNames as well. Unfortunately a unit test for `toGoStrings` was not added since "use of cgo in test is not supported"


OmriTreidel:
@deadprogram Is there any chance someone will look at it? It's been pending for months. Also, there is something funny happening with the codecov check: even though I've covered 100% of the code I've changed the overall coverage has dropped.

JoshChristie (approved):
Looks great.

deadprogram:
https://github.com/hybridgroup/gocv/pull/732 fixed the issue with `GetLayerNames()`.

Perhaps you can reduce this PR back to just adding the functions mentioned in #674 which would make it a lot easier to review.

Thanks!

OmriTreidel:
@deadprogram The function `toGoStrings` is needed for decoding the QR codes as part of #674. The usage of this function in `GetLayerNames` is actually reducing the size of the code base. I can revert it but I don't think it simplifies the review at all.

deadprogram:
It would have made it a lot easier for me to review more quickly if there had been a PR adding `toGoStrings()` and then a separate one adding `detectAndDecodeMulti()`. However, now I understand what it is doing, seems like a reasonable approach.

Now going to merge, thanks for the contributions and sorry to have taken so long to get it reviewed.

PR #689, opened 2020-06-05
Change constants and corresponding function signatures to have the correct types

### LINKED ISSUE #686: Constants defined in const() block have the wrong type
## Description
Example: https://github.com/hybridgroup/gocv/blob/v0.23.0/imgcodecs.go#L13
```
// IMReadFlag is one of the valid flags to use for the IMRead function.
type IMReadFlag int

const (
	// IMReadUnchanged return the loaded image as is (with alpha channel,
	// otherwise it gets cropped).
	IMReadUnchanged IMReadFlag = -1

	// IMReadGrayScale always convert image to the single channel
	// grayscale image.
	IMReadGrayScale = 0
```
`IMReadUnchanged` has type `IMReadFlag`.
`IMReadGrayScale` has type `int` because a value is specified and the type annotation is not repeated.

## Steps to Reproduce
See https://play.golang.org/p/wYFVwKAFicK for a minimal example.

Minnozz:
Annotate constant in const() blocks with the correct type if they have a value specified. This also changes some function parameter types from `int` to the correct type.

Fixes #686.

### DIFF
--- a/core.go
+++ b/core.go
@@ -959,7 +959,7 @@ func BitwiseXorWithMask(src1 Mat, src2 Mat, dst *Mat, mask Mat) {
 // For further details, please see:
 // https://docs.opencv.org/master/d2/de8/group__core__array.html#ga4ba778a1c57f83233b1d851c83f5a622
 //
-func BatchDistance(src1 Mat, src2 Mat, dist Mat, dtype int, nidx Mat, normType int, K int, mask Mat, update int, crosscheck bool) {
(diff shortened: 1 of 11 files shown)


deadprogram:
Thanks for doing this @Minnozz now squash/merging.

Minnozz:
@deadprogram No problem.

Just to be sure (I maybe should have been clearer about this in my PR): this change may require users to make (small) changes to their code when they update to this version. The change in the tests is an example of this. I don't know what the backwards compatibility promise of GoCV is, but you may want to mention this in the changelog when a new version is released.

PR #697, opened 2020-06-12
Add MultiplyWithParams

### LINKED ISSUE #696: Add MultiplyWithParams()
## Description
The current [Multiply](https://pkg.go.dev/gocv.io/x/gocv?tab=doc#Multiply) has no `scale` (and `dtype`) parameters, like [the OpenCV function](https://docs.opencv.org/master/d2/de8/group__core__array.html#ga979d898a58d7f61c53003e162e7ad89f) does. I'll make a PR that adds `MultiplyWithParams(src1 Mat, src2 Mat, dst *Mat, scale float64, dtype MatType)`, since that seems to fit into the current naming scheme best.

An alternative would be adding `MultiplyScale(src1 Mat, src2 Mat, dst *Mat, scale float64)`, that adds `scale` but ignores `dtype`.

Minnozz:
Closes #696.

### DIFF
--- a/core.cpp
+++ b/core.cpp
@@ -574,6 +574,10 @@ void Mat_Multiply(Mat src1, Mat src2, Mat dst) {
     cv::multiply(*src1, *src2, *dst);
 }
 
+void Mat_MultiplyWithParams(Mat src1, Mat src2, Mat dst, double scale, int dtype) {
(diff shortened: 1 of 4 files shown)


deadprogram:
Hi @Minnozz thank you very much for this contribution. Now merging.

PR #710, opened 2020-07-11
features2d: correct Blue and Red colours in gocv.DrawKeyPoints()

golubaca:
Changed color value order to reflect ones in other functions.


deadprogram:
Thank you for this fix @golubaca now merging.

PR #729, opened 2020-08-25
Added ability to load Torch file

mattnelsonuk:
Exactly the same as loading other DNN nets, but this one was missing. Allows you to load .t7 files.

```
embedder := gocv.ReadNetFromTorch(embeddingModel)
```

Tested using `openface.nn4.small2.v1.t7` to get 128d vectors from an image without any issues.


deadprogram:
Thanks @silverark for the contribution. Now merging.

PR #736, opened 2020-09-15
Add NMSBoxes function

freckie:
There's no Go binding of NMSBoxes function, I made it.
On my project, it works well.

Reference:
https://docs.opencv.org/4.4.0/d6/d0f/group__dnn.html#ga9d118d70a1659af729d01b10233213ee


deadprogram:
Hello @freckie first of all, thank you for another contribution.

Seems to me that the code should be added to the existing `dnn.go`, `dnn.cpp`, and `dnn.h` files, instead of adding new files.

Also, can you please add a smoke test for the new function?

Thanks!

freckie:
Hello @deadprogram !
I moved the code to `dnn.h`, `dnn.cpp`, `dnn.go`, and I also add two test functions named `TestNMSBoxes` and `TestNMSBoxesWithParams` to `dnn_test.go`.

deadprogram:
Thank you for the contribution and for making the requested changes @freckie 

Now going to squash/merge. Thanks again!

Issue #205, opened 2018-05-24
Memory leak in SIFT?


Lezh1k:
Hello there. 
I have one really strange problem with SIFT detector. I use it in our service and seems like there is memory leak. 
I use method DetectAndCompute. When I looked at code I found that `defer KeyPoints_Close` is commented. I was able to make new test binding KeyPoints_Close2 and free memory, but it didn't solve my problem. 
I have no clue how to reproduce it without using service. I mean simple application doesn't cause such big memory consumption. 

For example here is some code: 
```
func TestSift(t *testing.T) {
	gray := gocv.IMRead("path_to_image", gocv.IMReadGrayScale)

	sift := contrib.NewSIFT()
	defer sift.Close()

	res := make([]point, 0) //this is used as result in service's code.
	for iter := 0; iter < 50; iter++ {
		mask := gocv.NewMat()
		defer mask.Close()
		k, d := sift.DetectAndCompute(gray, mask)
		defer d.Close()

		for i, v := range k {
			var tmp []float64
			for j := 0; j < d.Cols(); j++ {
				tmp = append(tmp, float64(d.GetFloatAt(i, j)))
			}
			res = append(res, point{keypoint: v, descriptor: tmp})
		}
	}
	utils.ShowImage(gray) //This function just shows image in new window. 
}
```

Similar code is used in my service and I have crashes (because of memory problem). 
Interesting thing - using BRISK (for example) fixes this problem, but using SURF - not.

I wasn't able to find significant difference between BRISK and SIFT bindings, so maybe there is another problem. I have C++ application and it hasn't such problem, so maybe there is some golang specific things. 

I tried to use pprof, but without any luck (But maybe I didn't use it in right way). Valgrind also shows that there is no memory issues.

I just have no clue what's happening there. Top shows, that memory usage increases by ~300Mb after each call.  Does anybody face this problem?

tzununbekov:
It may be useful to have steps to reproduce:
1) `go run main.go`:
https://gist.github.com/tzununbekov/8c9573f75ea5a35d0c3c15b192adbc7e

2) `curl -i -X POST -H "Content-Type: multipart/form-data" -F "img=@<image>" http://localhost:8080/upload`

3) After a couple curl requests `top` (or any other analog) will show over 600Mb of resident memory allocated to this process while heap memory profile (inuse_space, alloc_space) usually shows not more than 50Mb

deadprogram:
[...]

PR #740, opened 2020-09-24
Fix a memory leak issue in SIFT.

forcey:
Hi,

I noticed a memory leak issue in the SIFT detection code path, where the C.KeyPoints structures are not closed after extraction. This may be related to https://github.com/hybridgroup/gocv/issues/205, but it was closed more than 2 years ago.

Here's a demo: https://pastebin.com/CTiE2SKJ. The image I'm using is https://upload.wikimedia.org/wikipedia/commons/c/c7/Caravaggio_-_Boy_Bitten_by_a_Lizard.jpg.

Before the fix, the process would use ~1G memory after 1280 iterations (this would take ~40 mins on my machine). After the fix it stays at ~30MB after GC periodically kicks in.

Environment:
macOS Catalina
go version go1.15 darwin/amd64

Thanks,
Todd

### DIFF
--- a/features2d.go
+++ b/features2d.go
@@ -779,6 +779,7 @@ func (d *SIFT) Close() error {
 //
 func (d *SIFT) Detect(src Mat) []KeyPoint {
 	ret := C.SIFT_Detect((C.SIFT)(d.p), C.Mat(src.Ptr()))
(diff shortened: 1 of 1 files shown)


deadprogram:
Thanks for finding this @forcey now rebase/merging.

PR #743, opened 2020-09-30
Add option to return baseline with GetTextSizeWithBaseline

Chekov2k:
Currently, there is no way to get the baseline value of the opencv getTextSize function. Added GetTextSizeWithBaseline and used it with GetTextSize. Let me know what you think


deadprogram:
Thank you for the contribution @Chekov2k I switched to the `dev` branch as mentioned here: https://github.com/hybridgroup/gocv/blob/release/CONTRIBUTING.md#how-to-use-our-github-repository

Now merging, thanks again!

Chekov2k:
Apologies for forking the wrong branch, should have read more carefully what to do. Thank you for fixing it and merging :)

PR #755, opened 2020-10-22
Keep a reference to a []byte that backs a Mat.

brunoga:
- This prevents crashes when Mats created with NewMatFromBytes are used.

### DIFF
--- a/core.go
+++ b/core.go
@@ -178,6 +178,9 @@ var ErrEmptyByteSlice = errors.New("empty byte array")
 //
 type Mat struct {
 	p C.Mat
+
+	// Non-nil if Mat was created with a []byte (using NewMatFromBytes()). Nil otherwise.
+	d []byte
 }
 
 // NewMat returns a new empty Mat.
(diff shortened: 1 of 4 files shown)


brunoga:
A few comments:

- This is the least surface-change version I could come up with, but there are possible different approaches (they will all boil down to something like this anyway.
- I added some TODOs. They can be removed if you prefer but I like to create them when I think of improvements/alternatives.
- Did not add tests but this can be tested so if you want point me to where the tests should live and I will do it.

deadprogram:
Hi @brunoga thanks for the contribution. I think leaving in the comments is a good thing, it helps explain decisions for possible future changes.

I changed this PR to the `dev` branch as mentioned in our contribution guidelines here: https://github.com/hybridgroup/gocv/blob/release/CONTRIBUTING.md#how-to-use-our-github-repository

Probably the best place to add a test would be next to this test here: https://github.com/hybridgroup/gocv/blob/dev/core_test.go#L24

Thanks!

brunoga:
I added a test that should catch the existing issue.

deadprogram:
@brunoga I am not seeing any test added to this branch. Might it be that it still needs to be pushed? Thanks.

brunoga:
Sorry for that, it is there now.

deadprogram:
Hi @brunoga looks like something not working as expected in Windows:
https://ci.appveyor.com/project/deadprogram/gocv/builds/36016481#L642

What do you think?

brunoga:
Sorry, this was an actual bug in the test (I did not close the Mat). Should be fixed now.

deadprogram:
Thanks for working on this @brunoga and making the requested changes. Now squash/merging.

PR #765, opened 2020-11-11
Add multidimensional Mat creation functions

JoshChristie:
This PR adds functions to create Mat's for creating multidimensional arrays.


deadprogram:
Thanks for another very useful PR @JoshChristie now merging.

Issue #773, opened 2020-12-08
Is anyone working on add cv::seamlessClone function in photo module


snowcrumble:
## Description
Is anyone working on add cv::seamlessClone function in photo module? If not, I'd like to do this

deadprogram:
Sounds great, please and thank you.

PR #775, opened 2020-12-09
photo: add SeamlessClone function

snowcrumble:
This PR adds `SeamlessClone` function, it's a wrapper of cv::seamlessClone.
Related issue: #773

### DIFF
--- a/cgo.go
+++ b/cgo.go
@@ -8,6 +8,6 @@ package gocv
 #cgo !windows pkg-config: opencv4
 #cgo CXXFLAGS:   --std=c++11
 #cgo windows  CPPFLAGS:   -IC:/opencv/build/install/include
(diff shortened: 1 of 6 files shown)


snowcrumble:
@deadprogram  excuse me, the CI fails and looks like there is no photo module in the CI environments. It's OK on my env, I use MacOS and installed opencv from brew, the header file `<opencv2/opencv_modules.hpp>` defines HAVE_OPENCV_PHOTO macro.

deadprogram:
You need to add that module specifically on Windows, due to not having `pkg-config`.

Please see https://github.com/hybridgroup/gocv/blob/release/cgo.go#L11

deadprogram:
Looks great @snowcrumble thank you very much for the contribution. Now merging.

Issue #771, opened 2020-12-05
matprofile does not track Split() mats


squizzling:
When matprofile is enabled, `gocv.Split()` does not track the creation of mats.

## Description
Because `gocv.Split()` allocates a `[]Mat` and sets each `Mat.p` [directly](https://github.com/hybridgroup/gocv/blob/release/core.go#L1784-L1786), it bypasses the profiling code.

A similar pattern appears to be used in [`Net.ForwardLayers`](https://github.com/hybridgroup/gocv/blob/release/dnn.go#L180-L183) and [`ImagesFromBlob`](https://github.com/hybridgroup/gocv/blob/release/dnn.go#L373)

`ImagesFromBlob` in particular looks like it might be a risk, because if the mats passed in the slice have a `p`, then it won't be released.

I have no direct knowledge of either of those last two being genuine problems, I just looked for writes to `Mat.p` in Goland.

## Steps to Reproduce
```
package main

import (
        "bytes"
        "fmt"

        "gocv.io/x/gocv"
)

func PrintProfile(message string) {
        fmt.Printf(message, gocv.MatProfile.Count())
        b := bytes.Buffer{}
        gocv.MatProfile.WriteTo(&b, 1)
        fmt.Print(b.String())
}

func CloseMat(mat gocv.Mat) {
        err := mat.Close()
        if err != nil {
                panic(err)
        }
}

func main() {
        mat := gocv.NewMatWithSize(5, 5, gocv.MatTypeCV8UC3)
        PrintProfile("count after creating 3 channel mat is %d\n")

        channels := gocv.Split(mat)
        PrintProfile("count after split is %d\n")

        for _, channel := range channels {
                CloseMat(channel)
        }
        PrintProfile("count after closing channels is %d\n")

        CloseMat(mat)
        PrintProfile("count after closing 3 channel mat is %d\n")
}
```

## Your Environment
Tested in hybridgroup/gocv@sha256:1a48836ce6b31adaaf45f5b85210d6f43e1eeadfd2d4c535c86d97beee42593b (`latest` at time of writing)

golubaca:
I can confirm that this is a bug. I already fixed this issue, as soon as tests are finished, I'll create a pull request. Thanks for reporting this issue.

golubaca:
All tests passed, because of this issue many tests were corrected because of unclosed mats returned by `Split` function.

deadprogram:
This was released with v0.26.0 so now closing. Thank you everyone!

PR #780, opened 2020-12-12
Add created mats in Split and ForwardLayers to profile

golubaca:
This change reflect issue #771 where mats created inside Split and ForwardLayers  are not registered inside profile. After creating new function which just adds passed mats to profile, those mats are properly registered.


golubaca:
Ok, so many unclosed Mats in many tests are forcing build to fail. I will update this PR after closing all unclosed mats in tests.

deadprogram:
Thank you very much for this contribution @golubaca I am going to squash/merge now. Thanks again!

Issue #657, opened 2020-04-11
Mat.ToImage() always returns *image.RGBA, even if mat is YUV


MikolajMGT:
Hello there!
In current implementation Mat.ToImage() method always returns *image.RGBA, which may be not expected behaviour. Consider following code
```
dev.Read(&mat)
gocv.CvtColor(mat, &yMat, gocv.ColorRGBToYCrCb)
img, err := yMat.ToImage()
ycbcr := img.(*image.YCbCr)
```
It won't work, beacuse of `panic: interface conversion: image.Image is *image.RGBA, not *image.YCbCr
`
Sometimes it is necessary to work with *image.YcbCr and there is no simple way to make rgb->ycbcr conversion in Go, so I would be glad if such functionality would be provided :)
Greetings

deadprogram:
This is going to require a new function like `ToImageYUV()` to perform the correct conversion.

golubaca:
I have a question, should we go easy way or build maybe optimized version? If we take a shortcut and use `ToImage` function, which we translate to YUV, we get simplicity, but don't know about performance. Maybe something like this:
```golang
func (m *Mat) ToImageYUV() (*image.YCbCr, error) {
	img, err := m.ToImage() //using ToImage function to get RGBA
	if err != nil {
		return nil, err
	}
	bounds := img.Bounds()
	converted := image.NewYCbCr(bounds, image.YCbCrSubsampleRatio420)
        
        // creating YUV image from RGB
	for row := 0; row < bounds.Max.Y; row++ {
		for col := 0; col < bounds.Max.X; col++ {
			r, g, b, _ := img.At(col, row).RGBA()
			y, cb, cr := color.RGBToYCbCr(uint8(r), uint8(g), uint8(b))

			converted.Y[converted.YOffset(col, row)] = y
			converted.Cb[converted.COffset(col, row)] = cb
			converted.Cr[converted.COffset(col, row)] = cr
		}
	}
	return converted, nil
}
```

And we can use it like this:
```golang
package main

import (
	"image/jpeg"
	"log"
	"os"

	"gocv.io/x/gocv"
)

func main() {
	mat := gocv.IMRead("input.jpg", 1)
	img, err := mat.ToImageYUV()
	if err != nil {
		panic(err)
	}
	f, err := os.Create("output.jpg")
	if err != nil {
		panic(err)
	}
	defer f.Close()
	jpeg.Encode(f, img, nil)
}
```

PR #781, opened 2020-12-14
imgproc: Add ToImageYUV and ToImageYUVWithParams

golubaca:
Added conversion from image.RGBA to image.YCbCr based on #657

### DIFF
--- a/imgproc.go
+++ b/imgproc.go
@@ -1994,6 +1994,50 @@ func (m *Mat) ToImage() (image.Image, error) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


deadprogram:
Hi @golubaca just was looking over this PR. I wonder if first using `CvtColor()` to do the color space conversion, and then copying the resulting data directly into the `Y, Cb, Cr` slices of an `image.YCbCr` might be another way to do this. What do you think?

golubaca:
Hi @deadprogram I tried many solutions, but some failed and some was much slower that this one. Problem with YUV is that is not that simple as RGB and sometimes I got some weird darker image.

deadprogram:
OK, time to merge this already. Thanks for working on it @golubaca

Issue #787, opened 2020-12-23
Support LineType in gocv.Circle, gocv.Rectangle and gocv.FillPoly


renewboy:
The `LineType` is very useful to render the shapes with antialiasing. Is there any plan to support this parameter?

golubaca:
Sure, it can be done with `FunctionWithParams` (e.g. `CircleWithParams(img *Mat, center image.Point, radius int, c color.RGBA, thickness, lineType, shift int )`). I can start working on this as soon as I can.

renewboy:
Thanks a million!!

mbalayil:
Please take a look at my PR.

deadprogram:
Released in v0.28 so now closing. Thanks!

PR #788, opened 2020-12-24
imgproc: Add WithParams functions to addres #787

golubaca:
Added:

- FillPollyWithParams
- CircleWithParams
- RectangleWithParams
- EllipseWithParams
- PolylinesWithParams

Added new and changed current tests to cover more cases.

### DIFF
--- a/imgproc.go
+++ b/imgproc.go
@@ -1273,7 +1273,7 @@ func Ellipse(img *Mat, center, axes image.Point, angle, startAngle, endAngle flo
 	C.Ellipse(img.p, pc, pa, C.double(angle), C.double(startAngle), C.double(endAngle), sColor, C.int(thickness))
 }
 
-// Ellipse draws a simple or thick elliptic arc or fills an ellipse sector.
+// EllipseWithParams draws a simple or thick elliptic arc or fills an ellipse sector.
 //
 // For further details, please see:
 // https://docs.opencv.org/master/d6/d6e/group__imgproc__draw.html#ga28b2267d35786f5f890ca167236cbc69
@@ -1497,6 +1497,47 @@ func Polylines(img *Mat, pts [][]image.Point, isClosed bool, c color.RGBA, thick
 	}
 
 	C.Polylines(img.p, cPoints, C.bool(isClosed), sColor, C.int(thickness))
+}
+
+// PolylinesWithParams draws several polygonal curves.
+//
+// For more information, see:
+// https://docs.opencv.org/master/d6/d6e/group__imgproc__draw.html#ga1ea127ffbbb7e0bfc4fd6fd2eb64263c
+func PolylinesWithParams(img *Mat, pts [][]image.Point, isClosed bool, c color.RGBA, thickness int, lineType LineType, shift int) {
+	points := make([]C.struct_Points, len(pts))
+
+	for i, pt := range pts {
+		p := (*C.struct_Point)(C.malloc(C.size_t(C.sizeof_struct_Point * len(pt))))
+		defer C.free(unsafe.Pointer(p))
+
+		pa := getPoints(p, len(pt))
+
+		for j, point := range pt {
+			pa[j] = C.struct_Point{
+				x: C.int(point.X),
+				y: C.int(point.Y),
+			}
+		}
+
+		points[i] = C.struct_Points{
+			points: (*C.Point)(p),
+			length: C.int(len(pt)),
+		}
+	}
+
+	cPoints := C.struct_Contours{
+		contours: (*C.struct_Points)(&points[0]),
+		length:   C.int(len(pts)),
+	}
+
+	sColor := C.struct_Scalar{
+		val1: C.double(c.B),
+		val2: C.double(c.G),
+		val3: C.double(c.R),
+		val4: C.double(c.A),
+	}
+
+	C.PolylinesWithParams(img.p, cPoints, C.bool(isClosed), sColor, C.int(thickness), C.int(lineType), C.int(shift))
 }
 
 // HersheyFont are the font libraries included in OpenCV.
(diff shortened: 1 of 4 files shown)


imgproc.go
```diff
@@ -1343,6 +1413,51 @@ func FillPoly(img *Mat, pts [][]image.Point, c color.RGBA) {
...
+func FillPolyWithParams(img *Mat, pts [][]image.Point, c color.RGBA, lineType LineType, shift int, offset image.Point) {
+	points := make([]C.struct_Points, len(pts))
+	offsetP := C.struct_Point{
+		x: C.int(offset.X),
+		y: C.int(offset.Y),
+	}
+
+	for i, pt := range pts {
+		p := (*C.struct_Point)(C.malloc(C.size_t(C.sizeof_struct_Point * len(pt))))
+		defer C.free(unsafe.Pointer(p))
```
deadprogram: I do not think you want to use `defer` here. Instead I suggest explicitly freeing `p` at the end of each loop iteration.

imgproc.go
```diff
@@ -1384,6 +1499,47 @@ func Polylines(img *Mat, pts [][]image.Point, isClosed bool, c color.RGBA, thick
...
+// PolylinesWithParams draws several polygonal curves.
+//
+// For more information, see:
+// https://docs.opencv.org/master/d6/d6e/group__imgproc__draw.html#ga1ea127ffbbb7e0bfc4fd6fd2eb64263c
+func PolylinesWithParams(img *Mat, pts [][]image.Point, isClosed bool, c color.RGBA, thickness int, lineType LineType, shift int) {
+	points := make([]C.struct_Points, len(pts))
+
+	for i, pt := range pts {
+		p := (*C.struct_Point)(C.malloc(C.size_t(C.sizeof_struct_Point * len(pt))))
+		defer C.free(unsafe.Pointer(p))
```
deadprogram: See previous comment.

imgproc.go
golubaca: Make sense. I just used existing working code for `WithParams` functions, so old code needs to be refactored as well.

deadprogram: No doubt an audit of the code will reveal a few more of these to be fixed.

golubaca:
checking circleci, failed because of `dl.caffe.berkeleyvision.org` which is currently unavailable.

deadprogram:
Hello @golubaca the caffee download site is reachable again, so I restarted the build. Now there are some actual errors appearing in the tests themselves.

golubaca:
@deadprogram sometimes I get this same errors myself on my local machine. It is from gstreamer api PR. But they randomly appear and disappear.

deadprogram:
```
=== RUN   TestFillPoly
    imgproc_test.go:1823: TestFillPoly(): wrong pixel value = 0, want = 255
--- FAIL: TestFillPoly (0.00s)
=== RUN   TestFillPolyWithParams
=== RUN   TestFillPolyWithParams/FillPollyWithParams
=== RUN   TestFillPolyWithParams/FillPollyWithParams_offset_2
--- PASS: TestFillPolyWithParams (0.00s)
    --- PASS: TestFillPolyWithParams/FillPollyWithParams (0.00s)
    --- PASS: TestFillPolyWithParams/FillPollyWithParams_offset_2 (0.00s)
=== RUN   TestPolylines
=== RUN   TestPolylines/Polylines
    imgproc_test.go:1888: TestPolylines(): wrong pixel value = 0, want = 255
=== RUN   TestPolylines/PolylinesWithParams
    imgproc_test.go:1908: TestPolylines(): wrong pixel value = 0, want = 255
--- FAIL: TestPolylines (0.00s)
    --- FAIL: TestPolylines/Polylines (0.00s)
    --- FAIL: TestPolylines/PolylinesWithParams (0.00s)
```

Seems consistently failing on CircleCI.

golubaca:
I checked on multiple machines and all test passed. Appveyor pass tests as well. It can be connected to a docker image. I will test this case locally.

Checked with docker, and there is an failed test as well, so there is some difference with docker and custom build. Values are not the same. Easiest thing will be to revert to old tests which didn't check values, but we need to check more cases which will make this library more stable, so another solution is required. Same code is executed on windows build and docker image:
Docker:
From docker image
Win build:
from win build

That's why CI failed.

This can be confirmed by executing this test code on my forked branch:
```golang
package main

import (
	"image"
	"image/color"

	"gocv.io/x/gocv"
)

func main() {
	img := gocv.NewMatWithSize(100, 100, gocv.MatTypeCV8UC1)
	defer img.Close()

	white := color.RGBA{255, 255, 255, 0}
	pts := [][]image.Point{
		{
			image.Pt(10, 10),
			image.Pt(10, 20),
			image.Pt(20, 20),
			image.Pt(20, 10),
		},
	}
	gocv.FillPolyWithParams(&img, pts, white, gocv.Line4, 0, image.Point{})
	gocv.IMWrite("testBuild.jpg", img)
}
```
I checked many functions inside docker container and results follows this pattern, many drawings are off or different shape than expected, but many tests doesn't check for pixel values after drawing, so tests are passing.

deadprogram:
Hmm, we will need to delve into testing specific values in a future set of PRs. 

Although, it is true that some functions are OS specific (aka drawing to screen) vs. algorithms (should be the same on all platforms, with at most minor differences in precision).

In the meantime, what do you think would be the best way to proceed to get the CI passing on this PR?

golubaca:
Well, tests can be rewritten as before, to just check if function is working without checking for pixel values, and ci will pass.

deadprogram:
> Well, tests can be rewritten as before, to just check if function is working without checking for pixel values, and ci will pass.

That is probably our best option for the moment.

deadprogram:
Hello @golubaca I would love to try to revive this PR and help complete it.

There is now a merge conflict of course. But we also need to decide what to do about the tests themselves. Any further thoughts?

deadprogram:
Closing because this was implemented by #841

PR #790, opened 2020-12-28
Adding gcc-c++ package to rpm deps

FloThinksPi:
A clean fedora workstation does not come with a c++ compiler so we need to install that as a dep.


deadprogram:
Hello @FloThinksPi thank you for the PR. I moved the branch over to `dev` as mentioned in our contribution guidelines here: https://github.com/hybridgroup/gocv/blob/release/CONTRIBUTING.md#how-to-use-our-github-repository

Now merging, thanks again for the contribution!

PR #791, opened 2020-12-28
fix dldt repo in makefile for openvino

FloThinksPi:
The dltd repo does not exist anymore and is a redirect to https://github.com/openvinotoolkit/openvino.
Thus the branch does not exist there. I found 2019_R3.1 may be an adequat equivalent.

I dont know which rev the original branch 2019 was pointing at so maybe someone that used it before can check this.
Also test if this works not just on my machine :>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,14 +60,15 @@ download:
 	rm opencv.zip opencv_contrib.zip
 	cd -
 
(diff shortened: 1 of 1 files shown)


deadprogram:
Seems correct, so now merging. Thanks for the contribution @FloThinksPi

PR #795, opened 2021-01-14
add rand functions

snowcrumble:
This PR add some rand functions, and fix code style for RNG


deadprogram:
Thank you very much for another random contribution @snowcrumble (pun alert)

Now merging, thanks again!

PR #815, opened 2021-03-21
Oenvino with CUDA

alaayousef87:
I believe gocv support openvino with cuda in intel gpu plugin cldnn but in my case I cannot have the cldnn because I don't have intel GPU I have only Nvidia gtg 1060 geoforce 

I hope gocv they make option to build openvino opencv with CUDA in the cpu plugin of intel not the gpu

because my target is to run openvino in backend of NVidia cuda   

Please please let me know if I can do that with the current version of gocv???

I try a lot to change the build of openvino with the interface engine mkl_dnn "the cpu one" and I sucessfully build it from source with enable the mkl_dnn and disable cldnn but then failed with gocv,  because it is look that not possible in gocv because gocv will support only the cldnn

Thank you.


deadprogram:
Sorry, what exactly is it that you are trying to add in this pull request? Perhaps you meant to create an issue to ask a question.

deadprogram:
I see you did create an issue, so now closing this.

PR #816, opened 2021-03-25
use enum instead of int for Invert Method

snowcrumble:
flags parameter of Invert is same as method Solve's.


deadprogram:
Good catch @snowcrumble and thanks for the contribution. Now merging.

PR #823, opened 2021-04-05
Add legacy

Danile71:


deadprogram:
@Danile71 could you please add some smoke tests now for the legacy functions? Thanks!

PR #828, opened 2021-04-22
make: add Makefile tasks for OpenCV install on Nvidia Jetson

deadprogram:
This PR adds Makefile tasks to facilitate installing OpenCV 4.5.2 with CUDA on the Nvidia Jetson Nano developer kit.


deadprogram:
It worked for me so now merging.

PR #830, opened 2021-04-23
cuda: initial implementation for the Flip function

deadprogram:
This PR provides an implementation for the CUDA Flip function. It includes the code, as well as a test that shows that the code is able to be executed.

image


deadprogram:
Merging because I need this function now! :smile_cat:

PR #844, opened 2021-05-18
Add Xphoto contrib

Syd76:
Xphoto contrib :

On TESTING for Inpaint functions, there an 

>fatal error: unexpected signal during runtime execution
>[signal SIGSEGV , segmentation vialotion code=0x1 addr=0X01.... ]

But, an application run well ( see : cmd/xphoto/main.go)

### DIFF
--- a/cmd/xphoto/main.go
+++ b/cmd/xphoto/main.go
@@ -1,0 +1,120 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"gocv.io/x/gocv"
+	"gocv.io/x/gocv/contrib"
+	"os"
+)
+
+func Inpaint() {
+
+	jpgImage := gocv.IMRead("./images/space_shuttle.jpg", gocv.IMReadColor)
+
+	if jpgImage.Empty() {
+		fmt.Printf("Invalid read of Source Mat in TestInpaint test\n")
+		return
+	}
+
+	src := gocv.NewMat()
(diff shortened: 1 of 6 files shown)


deadprogram:
Hello @Syd76 thank you very much for this contribution.

I think you need to add `-lopencv_xphoto452` to the list of libs linked by Windows here: https://github.com/hybridgroup/gocv/blob/dev/contrib/cgo.go#L11

Since Windows does not have pkg-config support generally, all libs need to be in this list to be correctly linked.

Syd76:
> I think you need to add `-lopencv_xphoto452` to the list of libs linked by Windows here: https://github.com/hybridgroup/gocv/blob/dev/contrib/cgo.go#L11

Yep, sorry i didn't see this wrong AppVeyor result.
I work only under Linux/ubuntu.

Syd76:
[79f5a35](https://github.com/hybridgroup/gocv/pull/844/commits/79f5a35525c448f2ea90557bdf43fd1fa38f2c53)

- Change in order to be consistent with the Opencv function (parameter by reference)

deadprogram:
Thanks for the changes @Syd76 I think it is looking good.

Would you mind squashing the commits so I can merge? Or I can do it if you prefer.

deadprogram:
I will squash/commit now. Thank you very much for the awesome contribution @Syd76

PR #850, opened 2021-05-27
fix: change MOSSE to KCF

maxsxu:
MOSSE has been moved to legacy since OpenCV 4.5.1


deadprogram:
Good point @jsonbruce now merging. Thanks for the correction.

Issue #851, opened 2021-05-31
panic: pprof: Profile.Add of duplicate value


rowdyroad:
Panic on mat.Clone with matprofile tag

## Description
panic: pprof: Profile.Add of duplicate value

goroutine 20 [running]:
runtime/pprof.(*Profile).Add(0xc000078a80, 0xca11a0, 0x7f9001d4c900, 0x1)
	/usr/local/go/src/runtime/pprof/pprof.go:298 +0x295
gocv.io/x/gocv.newMat(...)
	/go/pkg/mod/gocv.io/x/gocv@v0.23.0/mat_profile.go:64
gocv.io/x/gocv.(*Mat).Clone(0xc0001b5e90, 0xc0001b5c90)
	/go/pkg/mod/gocv.io/x/gocv@v0.23.0/core.go:245 +0x5f
go-face/pkg/apps/catcher/components.(*Scoper).Process(0xc0000a62c0, 0xc000330000, 0xc0001b5e90, 0xa9, 0x15d, 0x97f, 0x5ac, 0xc0001b5ea8, 0xc000414720)
	/app/pkg/apps/catcher/components/scoper.go:74 +0x6e

`srcClone := src.Clone()`

## Steps to Reproduce
1. Build with  -tags matprofile
2. Clone of gocv.Mat make panic (sometimes, not every call)
3. W/o  -tags matprofile everything's fine

## Your Environment

* Operating System and version: macos/linux
* OpenCV version used: 4
* How did you install OpenCV? docker gocv/opencv
* GoCV version used: 0.23.0
* Go version: 1.15
* Did you run the `env.sh` or `env.cmd` script before trying to `go run` or `go build`?

jheidel:
For what it's worth, I was getting about one `Add of duplicate value` panic a day using `-matprofile`. With my fix 4c86e5f, I haven't seen any panics so far. My samples sizes are a bit small: 3 days before, 2 days after.

rowdyroad:
Good job. I’m goin to check. Thanks 🖖

deadprogram:
I think this can now be closed, since #854 was merged. Please reopen if needed. Thanks!

PR #854, opened 2021-06-07
Fix possible MatProfile race by ordering remove before free.

jheidel:
When building with `-tags matprofile` set, gocv will occasionally panic due to a
duplicate pointer added to the pprof Profile: #851 

The pointer added comes directly from new (via Mat_New). Assuming correct memory
allocator behavior, that means that we must be inserting a pointer which was
previously deleted.

Since the pointer is currently removed from the Profile *after* it is freed,
there's a possible data race in multithreaded applications where `new` gives us
back a deleted pointer immediately and we insert it before we have a chance to
remove it from the profile, triggering this condition.

I'm not completely certain this resolves the issue since the race is extremely
difficult to reproduce.

### DIFF
--- a/mat_profile.go
+++ b/mat_profile.go
@@ -73,8 +73,10 @@ func newMat(p C.Mat) Mat {
 
 // Close the Mat object.
 func (m *Mat) Close() error {
-	C.Mat_Close(m.p)
+	// NOTE: The pointer must be removed from the profile before it is deleted to
+	// avoid a data race.
 	MatProfile.Remove(m.p)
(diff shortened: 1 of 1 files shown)


deadprogram:
Hello @jheidel I do not think you meant to include commit d105807 in this PR did you? Perhaps you can remove it please.

jheidel:
Removed it, sorry about that.

deadprogram:
Seems like a very good idea. Now merging, thanks @jheidel

PR #866, opened 2021-06-27
imencode: optimize IMEncode avoiding multiple data copies.

solganik:
Currently IMEncode implemented the following way:
1) In Image_IMEncode std::vector is allocated and filled by cv::imencode
2) In Image_IMEncode function toByteArray allocated another native array
and copies the data from the vector
3) In IMEncode the result of Image_IMEncode is further copied to golang
allocated byte array.

Hence there are 2 additional copies of the data for each call to
IMEncode.
This is highly inefficient, especially if this needs to be performed
frequently.

The solution is to avoid copies altogether (though with a slight API
cost).
First we introduce NativeByteBuffer struct which is just a wrapper for
std::vector. Note that it is hard-coded 24 bytes (std::vector is 3
pointers of data, and this will never change as this will force
recompilation of probably every existing c++ application).

Second We call Image_IMEncode with this "preallocated" and
pre-initialied std::vector.

Finally the "func (buffer *NativeByteBuffer) GetBytes() []byte" function
returns a hand-made slice from the underlying std::vector buffer.

Note that we both provide a "Close" function and register "finalizer" to
clear the native buffer when NativeByteBuffer goes out of scope.

Usage Example:

buf, err := IMEncode(PNGFileExt, img)
defer buf.Close()
bytes := buf.GetBytes()

### DIFF
--- a/cmd/mjpeg-streamer/main.go
+++ b/cmd/mjpeg-streamer/main.go
@@ -80,6 +80,7 @@ func mjpegCapture() {
 		}
 
 		buf, _ := gocv.IMEncode(".jpg", img)
-		stream.UpdateJPEG(buf)
+		stream.UpdateJPEG(buf.GetBytes())
+		buf.Close()
 	}
 }
--- a/core.cpp
+++ b/core.cpp
@@ -1010,3 +1010,19 @@ void copyPointVectorToPoint2fVector(PointVector src, Point2fVector dest) {
         dest->push_back(cv::Point2f(src->at(i).x, src->at(i).y));
     }
 }
+
+void StdByteVectorInitialize(void* data) {
+    new (data) std::vector<uchar>();
+}
+
+void StdByteVectorFree(void *data) {
+    reinterpret_cast<std::vector<uchar> *>(data)->~vector<uchar>();
+}
+
+size_t StdByteVectorLen(void *data) {
+    return reinterpret_cast<std::vector<uchar> *>(data)->size();
+}
+
+uint8_t* StdByteVectorData(void *data) {
+    return reinterpret_cast<std::vector<uchar> *>(data)->data();
+}
--- a/core.go
+++ b/core.go
@@ -2534,3 +2534,43 @@ func RandU(mat *Mat, low, high Scalar) {
 
 	C.RandU(mat.p, lowVal, highVal)
 }
+
+type NativeByteBuffer struct {
+	// std::vector is build of 3 pointers And this will not change ever.
+	stdVectorOpaq [3]uintptr
+}
+
+func newNativeByteBuffer() *NativeByteBuffer {
+	buffer := &NativeByteBuffer{}
+	C.StdByteVectorInitialize(buffer.nativePointer())
+	return buffer
+}
+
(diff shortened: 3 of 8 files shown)


deadprogram:
Hi @solganik thanks for the contribution. Looks like you need to run `go fmt` on the changed files in order to pass the CircleCI tests. Could you please do that?

solganik:
> Hi @solganik thanks for the contribution. Looks like you need to run `go fmt` on the changed files in order to pass the CircleCI tests. Could you please do that?

done

deadprogram:
I think you have the right idea overall about avoiding copies. However, there are some issues with this implementation. First and foremost, we never use finalizers in GoCV. Since we cannot control when the GC will take place, instead we close `Mat` explicitly.

Also I think that it would be possible to reuse or change implementation of `ByteArray` to avoid creating another type, and at the same time still avoiding the extra copying. I have not looked closely at the implementation to do this, but I recall thinking about this at some point. See the `Mat.DataPtrUint8()` as a somewhat similar example.

deadprogram:
OK was looking at this further. I think if you just remove the finalizer parts, this is good. Maybe it would be better after that to make `ByteArray` more like the code in this PR? 

In any case, seems like a very good optimization at a fairly reasonable cost in API breakage.

solganik:
> OK was looking at this further. I think if you just remove the finalizer parts, this is good. Maybe it would be better after that to make `ByteArray` more like the code in this PR?
> 
> In any case, seems like a very good optimization at a fairly reasonable cost in API breakage.

ok .. i`ll remove them. Out of curiosity why dont you use both "Close' method and finalyzer  in order to avoid native memory leak .

solganik:
> > OK was looking at this further. I think if you just remove the finalizer parts, this is good. Maybe it would be better after that to make `ByteArray` more like the code in this PR?
> > In any case, seems like a very good optimization at a fairly reasonable cost in API breakage.
> 
> ok .. i`ll remove them. Out of curiosity why dont you use both "Close' method and finalyzer in order to avoid native memory leak .

done

deadprogram:
The short version to quote a member of the Go team:

> If you are using C code, you are using C's memory resource policy. So
the best approach is to use explicit calls to free. I agree that that
takes you back to the painful world of C's manual memory management,
but, after all, if you are using C then you can't pretend that you are
not using C.

Since we cannot control when the GC is run, we cannot count on the finalizer to free C memory. The problem with the dual approach is that you would need to avoid race conditions, and that drags in a bunch of other problems with trying to sync finalizer with the `Close()` calls.

Seems better to just have one explicit way to handle this.

solganik:
I dont quite understand the issue with races though. Finalizer will run
only if there are no references left. Which means that noone can call close
at this time, hence from concurrency perspective it is quite safe.
However i understand the unification reasoning and i removed the finalizer
from code.

deadprogram:
I think some investigation with pprof into the hybrid technique you propose seems warranted, ideally using `Mat`.

In any case, that is best done in a separate PR. 

Thanks for your efforts in this important area! Now merging this PR.

PR #802, opened 2021-02-07
all: make task and build tag for static build of OpenCV/GoCV on Linux

deadprogram:
This is a first attempt to provide a way to create a static build of OpenCV and consequent GoCV programs on Linux.

The info on how to do this was almost entirely figured out by @Danile71 I just am trying to put it into a more easily usable format.

This PR adds a make task and build tag that intended to be used as follows:

First, run the make task to build and install a static build of OpenCV:

***WARNING: this will replace your current shared build of OpenCV!***

```shell
make install_static
```

Once this is done, you should be able to build GoCV programs by using `-tags static` like this:

```shell
go build -tags static --ldflags '-extldflags "-static"' -o static_version ./cmd/version/main.go
```

The following warning is displayed:

```shell
/usr/local/lib/libopencv_videoio.a(backend_plugin.cpp.o): In function `cv::impl::PluginBackendFactory::loadPlugin()':
backend_plugin.cpp:(.text._ZN2cv4impl20PluginBackendFactory10loadPluginEv+0x120): warning: Using 'dlopen' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
```

Now you can run your statically linked program:

```shell
$ ./static_version 
gocv version: 0.27.0-dev
opencv lib version: 4.5.1
$ ls -l static_version 
-rwxr-xr-x 1 ron ron 61775512 feb  7 19:26 static_version
```

PR #870, opened 2021-07-02
Add static build abilitity to OpenCV/GoCV on Windows

puppywang:
Follow https://github.com/hybridgroup/gocv/pull/802, I add static build ability on windows. I modify the win_build_opencv.cmd file to add a optional argument to enable static build, when need to build static version of opencv, please use `win_build_opencv.cmd static`, then you can use `go build -tags static --ldflags '-extldflags="-static"' -o static_version.exe ./cmd/version/main.go` to create static version of your static-linked go executable on windows.


windosx:
Excellent work! It's very helpful.

PR #887, opened 2021-08-10
Update README.md to include export step for installation on macOS

kmans:
Adding in an important step to running GoCV on macOS. Without setting this environment variable, users running macOS may encounter an error similar to this one:

    
    Package opencv4 was not found in the pkg-config search path.
    Perhaps you should add the directory containing `opencv4.pc'
    to the PKG_CONFIG_PATH environment variable
    No package 'opencv4' found
    pkg-config: exit status 1

### DIFF
--- a/README.md
+++ b/README.md
@@ -367,7 +367,11 @@ You can install it by using Homebrew:
 
 To verify your installation you can run one of the included examples.
 
-First, change the current directory to the location of the GoCV repo:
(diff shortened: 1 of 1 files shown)


deadprogram:
Normally you would only have this problem if OpenCV has been installed as a "keg-only" package. Perhaps you needed to run `brew upgrade opencv` due to having multiple OpenCV installs already?

PR #894, opened 2021-08-30
build: use tbb for all builds for CPU accelerated operations

deadprogram:
This PR uses TBB (https://github.com/oneapi-src/oneTBB) for all builds in Makefile for CPU accelerated operations.


deadprogram:
Merging.

PR #895, opened 2021-09-02
Fix open video capture with api test

golubaca:
If tested with multiple devices connected, test fails because it should fail on deviceId 1, assuming that you have no more than 1 video device. Changing to 99 "assures" that in normal conditions test passes.


deadprogram:
Thanks for the updated test @golubaca do you want to squash your commits, or should I do it for you?

golubaca:
@deadprogram it would be nice if you can do it, I am on a vacation, so phone only available.

deadprogram:
Done. Thanks @golubaca

PR #926, opened 2022-01-25
openvino: Update env.sh and README.md

deradiri:
Decided to try out openvino and noticed the `env.sh` and `README.md` were a bit stale.

I was able to run the examples after updating `env.sh` and figured I'd open this PR since I couldn't find an existing one, or open issue.


deadprogram:
Thanks for updating this @deradiri now merging.

PR #927, opened 2022-01-25
dnn: Fix dangling pointers returned by `GetLayerNames`

deradiri:
`GetLayerNames` returns an array of `char` pointers to `cstrings` in a `vector<string>`; unfortunately, once the vector is out of scope, the strings are destroyed. `GetLayerNames` callers are then left with dangling pointers.

This change fixes the problem by expanding the `strs` buffer returned by `GetLayerNames` and copying the vector's cstrings into it.

**Testing**: 
_**See [this comment](https://github.com/hybridgroup/gocv/pull/927#issuecomment-1191250252)**_

Here's an example of the bug in action:
``` go

	n := net.GetLayerNames()

	for i, l := range n {
		fmt.Printf("Layer %d: %s\n", i, l)
	}
```

Before fix:
``` bash
$ go test -tags customenv,static
Layer 0: @�
Layer 1: 0�
...
```

After fix:
``` bash
$ go test -tags customenv,static
Layer 0: detector/yolo-v3-tiny/Conv_12/BiasAdd/YoloRegion
Layer 1: detector/yolo-v3-tiny/Conv_9/BiasAdd/YoloRegion
...
```

### DIFF
--- a/Dockerfile.opencv-openvino
+++ b/Dockerfile.opencv-openvino
@@ -5,10 +5,11 @@ LABEL maintainer="hybridgroup"
 ENV DEBIAN_FRONTEND=noninteractive
 USER root
 RUN apt-get update && apt-get install -y --no-install-recommends \
-            git build-essential cmake pkg-config unzip libgtk2.0-dev \
+            git build-essential cmake pkg-config unzip libgtk2.0-dev libgtk-3-0 \
             wget curl ca-certificates libcurl4-openssl-dev libssl-dev \
             libavcodec-dev libavformat-dev libswscale-dev libtbb2 libtbb-dev \
-            libjpeg-dev libpng-dev libtiff-dev libdc1394-22-dev && \
+            libjpeg-dev libpng-dev libtiff-dev libdc1394-22-dev libcanberra-gtk-module \
+            libcanberra-gtk3-module && \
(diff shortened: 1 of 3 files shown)


deadprogram:
Yes, I see how the previous implementation was lacking. I just wonder however how this test here has been passing? https://github.com/hybridgroup/gocv/blob/5bde20d79aa5c2376cb2ea16439252582321db28/dnn_test.go#L59

deradiri:
@deadprogram Good question... I guess maybe it's that the the caffe backend keeps the layer names around for longer/permanently?

I discovered this issue using openvino, will see if I can add a test case. Is there some procedure for adding models as test artifacts?

deradiri:
> Yes, I see how the previous implementation was lacking. I just wonder however how this test here has been passing?
> 
> https://github.com/hybridgroup/gocv/blob/5bde20d79aa5c2376cb2ea16439252582321db28/dnn_test.go#L59


@deadprogram It turns out this test passes by pure luck, I ran it a few times and logged all the layers. Below is a diff between two runs:

image

The layer name at index 1 seems to always be intact, so rather than add a new model I'll update the test to check a set of layers distributed throughout the array. Diff of runs with and without the fix in this PR is below:
image

Issue #930, opened 2022-02-05
gocv.Zeros() uses out of scope pointer


mitul93:
gocv.Zeros() function returns cv::Mat() object with `temp.data` which will be out of scope once the function returns. When elements of gocv.Mat object created with gocv.Zeros() is accessed, it will result in undefined behavior.

## Description
gocv.Zeros() return `cv::Mat(rows, cols, type, temp.data)` object where temp.data is data inside cv::Mat.
https://github.com/hybridgroup/gocv/blob/5730c4cf12d5fb4590fb7c97db8ed50c7020cc32/core.cpp#L68

However, the constructor only copies the pointer. It does not perform deep copy of the underlying data of `temp`
https://docs.opencv.org/4.x/d3/d63/classcv_1_1Mat.html#a51615ebf17a64c968df0bf49b4de6a3a

I suspect similar behavior for Eye() and Ones(). I haven't verified it.

## Steps to Reproduce
```
package main

import (
	"gocv.io/x/gocv"
)

func main() {
        // gocv.Mat with large enough data size to reproduce crash
	srcData := gocv.Zeros(1, 50000, gocv.MatTypeCV32FC1)

	for index := 0; index < srcData.Cols(); index++ {
		srcData.SetFloatAt(0, index, float32(10)) // setting data will cause crash
		//srcData.GetFloatAt(0, index) // Accessing data will cause crash
	}
}
```

`go run <filename>`

```
root@7058c3659c7f:/src/# go run crash_reproduce.go 
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0x7f1bb840d040 pc=0x4879e6]

runtime stack:
runtime.throw({0x4b3832, 0x3ff0000000000000})
        /usr/local/go/src/runtime/panic.go:1198 +0x71
runtime.sigpanic()
        /usr/local/go/src/runtime/signal_unix.go:719 +0x396

goroutine 1 [syscall]:
runtime.cgocall(0x481b50, 0xc00004bed0)
        /usr/local/go/src/runtime/cgocall.go:156 +0x5c fp=0xc00004bea8 sp=0xc00004be70 pc=0x41da1c
gocv.io/x/gocv._Cfunc_Mat_GetFloat(0x230e450, 0x0, 0x0)
        _cgo_gotypes.go:3905 +0x4d fp=0xc00004bed0 sp=0xc00004bea8 pc=0x480c2d
[...]

PR #931, opened 2022-02-07
core: correct how memory is being allocated for Eye(), Zeros(), and Ones() to address issue #930

deadprogram:
This PR corrects how memory is being allocated for Eye(), Zeros(), and Ones() to address issue #930


deadprogram:
Thanks for offline testing @mitul93 now merging.

PR #942, opened 2022-02-28
opencv version bump to 4.5.5

ksajan:


deadprogram:
Thanks @ksajan however please see https://github.com/hybridgroup/gocv/pull/948

PR #943, opened 2022-03-01
Fixed memory leaks in motion detection example

Visual-Vincent:
I noticed a memory leak in `cmd/motion-detect/main.go`: objects `kernel` and `contours` weren't being disposed of properly.


deadprogram:
Thanks for the correction @Visual-Vincent now merging.

PR #918, opened 2021-12-22
feat: add cv::wechat_qrcode::WeChatQRCode

dirname:
Reference: [cv::wechat_qrcode::WeChatQRCode](https://docs.opencv.org/4.x/d5/d04/classcv_1_1wechat__qrcode_1_1WeChatQRCode.html)

PR #949, opened 2022-03-10
feat: add cv::wechat_qrcode::WeChatQRCode

dirname:
fix: https://github.com/hybridgroup/gocv/pull/918#issuecomment-1062658077

@deadprogram


dirname (commented):
review: ready to merge

wechat_qrcode_test.go
```diff
@@ -0,0 +1,56 @@
+package gocv
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestNewWeChatQRCode(t *testing.T) {
```
deadprogram, line 9: Hello @dirname 

I am thinking that probably the tests and test model files should be handled conditionally like the test files for Caffe/Tensorflow. 

Please see https://github.com/hybridgroup/gocv/blob/release/dnn_test.go#L98

The reason for this is to keep the size of the repo down to a smaller size. 

Is there a place that the model files you are using for the tests here can be downloaded from, similar to how it is handled here for CircleCI: https://github.com/hybridgroup/gocv/blob/release/.circleci/config.yml#L25-L30 

and then here for Appveyor: https://github.com/hybridgroup/gocv/blob/release/appveyor_build_opencv.cmd#L19

dirname: @deadprogram Changed, please review again.

dirname (commented):
fix: https://github.com/hybridgroup/gocv/pull/949#discussion_r825261378

wechat_qrcode_test.go
deadprogram: Thanks for the quick update!

deadprogram:
@dirname do you mind squashing these commits together, or would you prefer that I do it before merge?

dirname:
@deadprogram You can choose the way you like, whether there is an estimated release time for the next release version, I want to be able to use the new features in the release version as soon as possible.

deadprogram:
I can squash and merge right now.

The next release is about to go out next week, so good timing!

dirname:
> I can squash and merge right now.
> 
> The next release is about to go out next week, so good timing!

Thanks for your awesome work, I will contribute as much as I can.

deadprogram:
I think I perhaps merged too soon. Just realized that this is part of `opencv_contrib` so really it needs to get moved to the `contrib` package inside of GoCV.

If you have time perhaps you can look at that please? Otherwise I will look into it this weekend.

dirname:
> I think I perhaps merged too soon. Just realized that this is part of `opencv_contrib` so really it needs to get moved to the `contrib` package inside of GoCV.
> 
> If you have time perhaps you can look at that please? Otherwise I will look into it this weekend.

https://github.com/hybridgroup/gocv/pull/952

PR #950, opened 2022-03-10
Update cgo_static.go

wgjtyu:
fix package name


deadprogram:
Thanks for the correction @wgjtyu now merging

PR #955, opened 2022-03-18
core: avoid index out of range panic in NewPointsVectorFromPoints

HearyShen:
In core.go, the `NewPointsVectorFromPoints` function has been implemented without validity check, it always accesses `points[0]` even if the `points` slice is empty. 
To avoid causing index out of range panic, I suggest returning a new empty `PointsVector` when input `pts` is empty.


deadprogram:
A most excellent suggestion @HearyShen thank you. Now merging.

PR #964, opened 2022-04-05
Add Mat.ElemSize

pancpp:
This method alongside of `Mat.Total` could help to tell the number of bytes occupied by the matrix.


deadprogram:
Thank you for the contribution @pancpp now squash/merging.

PR #997, opened 2022-07-03
photo: add inpaint function

RadhiFadlillah:
Currently `gocv` only has inpaint function from `xphoto` module in `opencv/contrib`. While it's good, it's too slow compared to the inpaint function from `photo` module. So, in this PR I've added binding to inpaint function in `photo` module, including its documentation and unit test.

By the way, I'm not really experienced using C language, so if I've made some mistake feel free to tell me and I'll fix it.

Thanks!


deadprogram:
Looks good to me, thank you for working on this @RadhiFadlillah now merging.

PR #999, opened 2022-07-05
Fix the problem that the non-maximum suppression return position is 0

Hexert:


Hexert:
image

For the same target, when the length of boxes and indices is 2, after the NMSBoxes method, the length of the returned indices is also 2, and the values in the indices are all 0. At this time, one of the boxes cannot be removed, and the appropriate target detection is retained. frame, so return the length of indicesVector to extract a suitable target detection frame

deadprogram:
Hello @Hexert thanks for looking into this.

Seems to me the correct solution is to set the size/values of the slice returned with `indices`, that way a developer does not have to "guess" the resulting size.

For example:

```go
...
indices = make([]int, indicesVector.length)
for i := 0; i < int(indicesVector.length); i++ {
		indices[i] = int(ptr[i])
}
return
```

Generally speaking we almost never add another param to a function signature that does not exist in the OpenCV function itself.

The same solution would also be needed for `NMSBoxesWithParams()`.

Also, the branch for PRs is normally against `dev`. See https://github.com/hybridgroup/gocv/blob/release/CONTRIBUTING.md for details.

Issue #1020, opened 2022-09-20
make deps outdated for ubuntu 22.04.01LTS


Hatsumi-FR:
Hey, I can't follow the documentation and install openCV, because libdc1394-22-dev package seems to be outdated since the new ubuntu release. It looks like the package has been remove https://askubuntu.com/questions/1407580/unable-to-locate-package-libdc1394-22-dev
## Description
```
hatsumi@hatsumi-X756UJ:~/src/gocv$ make deps
sudo apt-get -y update
sudo apt-get -y install unzip wget build-essential cmake curl git libgtk2.0-dev pkg-config libavcodec-dev libavformat-dev libswscale-dev libtbb2 libtbb-dev libjpeg-dev libpng-dev libtiff-dev libdc1394-22-dev
[sudo] Mot de passe de hatsumi : 
Atteint :1 http://fr.archive.ubuntu.com/ubuntu jammy InRelease
Atteint :2 http://fr.archive.ubuntu.com/ubuntu jammy-updates InRelease         
Atteint :3 http://fr.archive.ubuntu.com/ubuntu jammy-backports InRelease       
Atteint :4 https://download.docker.com/linux/ubuntu jammy InRelease            
Atteint :5 http://security.ubuntu.com/ubuntu jammy-security InRelease 
Lecture des listes de paquets... Fait
Lecture des listes de paquets... Fait
Construction de l'arbre des dépendances... Fait
Lecture des informations d'état... Fait      
E: Impossible de trouver le paquet libdc1394-22-dev
make: *** [Makefile:51 : deps_debian] Erreur 100
```

## Steps to Reproduce

1. Go to ubuntu 22.04.01LTS
2. Run the make deps command 
3.
4.

## Your Environment

* Operating System and version: Ubuntu 22.04.01LTS
* Go version: 1.18

dubstab96:
Same error while trying the installation guide. I have the same Ubuntu 22.04LTS with Go version 1.18

Hatsumi-FR:
@dubstab96 , I've found a solution to solve the problem until they fix it. In your makefile, you can replace "libdc1394-22-dev" by "libdc1394-dev" in DEBS and JETSON vars

dubstab96:
Thanks @Hatsumi-FR. That definitely worked :)

asbjorn:
This works for me too :+1: 
```
Description:	Ubuntu 22.04.1 LTS
Release:	22.04
Codename:	jammy
```

[...]

PR #1033, opened 2022-11-24
Changed the libdc1394 dependencies in Makefile for Ubuntu 22.04

asbjorn:
The original dependency (libdc1394-22) is outdated in Ubuntu 22.04 LTS. This fix just changes the dependency to be on the libdc1394 package itself, not the specific version.

This fix is from the github issue #1020


deadprogram:
Hello @asbjorn thank you for the code contribution. Could you please make this PR against the latest `dev` branch? See https://github.com/hybridgroup/gocv/blob/release/CONTRIBUTING.md#how-to-use-our-github-repository

asbjorn:
@deadprogram No problem - sorry for not reading the Contributing documentation first :)

deadprogram:
The main problem here is that the change is not backward compatible with the older LTS versions.

Some code to detect that, or else a separate make task, would solve that problem.

Also, this branch needs to be rebased against the latest `dev` branch.

Thanks!

spolti:
Was hit by this issue as well:
```
E: Package 'libtbb2' has no installation candidate
E: Unable to locate package libdc1394-22-dev
```

In addition:
```
Package libtbb2 is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  libtbbmalloc2
```

deadprogram:
Since the `Makefile` was refactored a while ago for various linux versions, this PR is no longer valid. Thanks however! Now closing.

PR #1035, opened 2022-11-24
imgcodecs: Add IMDecodeIntoMat to reduce heap allocations

NikeNano:
This PR adds [imdecode to mat](https://docs.opencv.org/4.x/d4/da8/group__imgcodecs.html#ga5a0acefe5cbe0a81e904e452ec7ca733) this function is overloaded and currently gocv only supports [new mat](https://docs.opencv.org/4.x/d4/da8/group__imgcodecs.html#ga26a67788faa58ade337f8d28ba0eb19e). 

I am completely 🆕  to `cgo` and believe I might have messed up here:

```
Mat Image_IMDecodeInToMat(ByteArray buf, int flags, Mat img) {
    std::vector<uchar> data(buf.data, buf.data + buf.length);
    cv::Mat imgOut = cv::imdecode(data, flags, img);
    return new cv::Mat(imgOut);
}
```

Will try to read up further but would be awesome with some help 🆘  on this. Thanks 🎉

### DIFF
--- a/imgcodecs.cpp
+++ b/imgcodecs.cpp
@@ -42,3 +42,8 @@ Mat Image_IMDecode(ByteArray buf, int flags) {
     cv::Mat img = cv::imdecode(data, flags);
     return new cv::Mat(img);
 }
+
+void Image_IMDecodeIntoMat(ByteArray buf, int flags, Mat dest) {
+    std::vector<uchar> data(buf.data, buf.data + buf.length);
+    cv::imdecode(data, flags, dest);
+}
--- a/imgcodecs.go
+++ b/imgcodecs.go
@@ -250,3 +250,19 @@ func IMDecode(buf []byte, flags IMReadFlag) (Mat, error) {
 	}
 	return newMat(C.Image_IMDecode(*data, C.int(flags))), nil
 }
+
+// IMDecodeIntoMat reads an image from a buffer in memory into a matrix.
+// The function IMDecodeIntoMat reads an image from the specified buffer in memory.
+// If the buffer is too short or contains invalid data, the function
+// returns an error
+//
+// For further details, please see:
+// https://docs.opencv.org/4.x/d4/da8/group__imgcodecs.html#ga5a0acefe5cbe0a81e904e452ec7ca733
+func IMDecodeIntoMat(buf []byte, flags IMReadFlag, dest *Mat) error {
+	data, err := toByteArray(buf)
+	if err != nil {
+		return err
+	}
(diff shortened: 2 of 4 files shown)


imgcodecs.go
```diff
@@ -242,11 +238,18 @@ func IMEncodeWithParams(fileExt FileExt, img Mat, params []int) (buf *NativeByte
...
-//
 func IMDecode(buf []byte, flags IMReadFlag) (Mat, error) {
 	data, err := toByteArray(buf)
 	if err != nil {
 		return Mat{}, err
 	}
 	return newMat(C.Image_IMDecode(*data, C.int(flags))), nil
 }
+
+func IMDecodeInToMat(buf []byte, flags IMReadFlag, mat Mat) (Mat, error) {
```
deadprogram: This function would be better named `IMDecodeIntoMat()` (note case)

Also, it needs to have GoDoc comments with description and a link to the OpenCV function it wraps like all the other functions, please.

deadprogram: Seems to me the signature of this function should be:

```go
func IMDecodeIntoMat(buf []byte, flags IMReadFlag, dest Mat) error
```

Since it is intended to put the result into `dest` instead of returning a new `Mat`.

Also, the implementation does not do this yet, although it needs to, correct? 😸

deadprogram: UPDATE: actually to conform to the function signatures used for other similar functions it should be

```go
func IMDecodeIntoMat(buf []byte, flags IMReadFlag, dest *Mat) error
```

imgcodecs.cpp
```diff
@@ -42,3 +42,8 @@ Mat Image_IMDecode(ByteArray buf, int flags) {
     cv::Mat img = cv::imdecode(data, flags);
     return new cv::Mat(img);
 }
+
+void Image_IMDecodeInToMat(ByteArray buf, int flags, Mat dest) {
```
NikeNano: Fix casing here!

imgcodecs_test.go
```diff
@@ -149,6 +149,25 @@ func TestIMDecode(t *testing.T) {
 		t.Error("Should not decode empty array")
 	}
 }
+
+func TestIMDecodeInToMat(t *testing.T) {
```
NikeNano: should be lower case.

NikeNano:
Are we ready to merge this @deadprogram  🤞 ?

deadprogram:
Thanks for the code contribution @NikeNano now squash/merging.

PR #1065, opened 2023-03-05
Update mistake url of cmd/README.md

ynikl:
Fix mistake url of  Facedetect from URL


deadprogram:
Those are actually 2 different example commands @ynikl not a mistake.

Issue #756, opened 2020-10-24
NMSBoxes out of bounds


photostorm:
## Description
When using newly added NMSBoxes function it seems that you have to create indices array with random value. If indices results from C.NMSBoxes returns more than the chosen size for indices array, then an out of bounds error will be throw.

## Steps to Reproduce
1. indices := make([]int, 1)
2. gocv.NMSBoxes(boxes, confidences, 0.5, 0.4, indices)
3. Will throw "panic: runtime error: index out of range [1] with length 1"

## Your Environment
gocv.io/x/gocv.NMSBoxes(0xc00014e400, 0x17, 0x20, 0xc0001b6200, 0x17, 0x20, 0x3ecccccd3f000000, 0xc000195e58, 0x1, 0x1)
        go/pkg/mod/gocv.io/x/gocv@v0.25.0/dnn.go:538 +0x386

* Operating System and version: Ubuntu 18.04
* OpenCV version used: 4.5.0
* How did you install OpenCV? from source with cuda
* GoCV version used: v0.25.0
* Go version: 1.15.3
* Did you run the `env.sh` or `env.cmd` script before trying to `go run` or `go build`?

deadprogram:
You might want to look at the test to see an example of this function is use: https://github.com/hybridgroup/gocv/blob/3a73f21769b6b8728df6a55f4f321315fc4539a3/dnn_test.go#L463

photostorm:
I did look at the test it. It does the same thing even if the array size is 10

deadprogram:
I would expect you need to create an array of exactly the correct size, from glancing at the implementation. Perhaps adding a `copy()` in the right place would be a good idea to ensure this result. We'd be happy to accept a PR that provided some additional checking.

PR #1068, opened 2023-03-23
Refactored NMSBoxes to return a slice, instead of supplying your own slice as a function parameter.

netbrain:
This seems like the more reasonable approach to this function call, Not sure why it was implemented this way in the first place.

The motivation behind this change is that you no longer have to invoke the code with the following: 

```
	nmsResult := make([]int, <some-arbitrary-number>)
	for i := range nmsResult {
		nmsResult[i] = -1
	}
	gocv.NMSBoxes(boxes, confidences, float32(0.45), float32(0.5), nmsResult)
	for _, idx := range nmsResult {
		if idx == -1 {
			break
		}
                // do something
	}
```

The new and improved code is now instead

```
	nmsResult := gocv.NMSBoxes(boxes, confidences, float32(0.45), float32(0.50))
	for _, idx := range nmsResult {
                // do something
         }

```

This is however a breaking change due to the signature change. So probably needs to be discussed.

### DIFF
--- a/dnn.go
+++ b/dnn.go
@@ -522,7 +522,7 @@ func (l *Layer) OutputNameToIndex(name string) int {
 //
 // For futher details, please see:
 // https://docs.opencv.org/4.4.0/d6/d0f/group__dnn.html#ga9d118d70a1659af729d01b10233213ee
-func NMSBoxes(bboxes []image.Rectangle, scores []float32, scoreThreshold float32, nmsThreshold float32, indices []int) {
+func NMSBoxes(bboxes []image.Rectangle, scores []float32, scoreThreshold float32, nmsThreshold float32) (indices []int) {
 	bboxesRectArr := []C.struct_Rect{}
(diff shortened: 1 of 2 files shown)


netbrain:
Relates to #756 and #736

netbrain:
@deadprogram do you have any input on this?

deadprogram:
I think the original motivation was to duplicate the C++ API, which takes a vector as param. The Python API does the same thing as what this PR proposes.

Since in either case we probably need to perform a heap allocation to return the results, it probably doesn't matter. Certainly this change makes it a lot nicer Go API.

As far as breaking change, that is a little unfortunate. The result is a simplified API however, so for me is probably worth it.

Anyone else using this function now and who wants to expression any thoughts on this proposed change?

netbrain:
@deadprogram time to merge?

deadprogram:
Switched to `dev` branch and now merging @netbrain thanks for working on this.

PR #1078, opened 2023-05-14
Ubuntu Jammy (22) opencv4 build support.

mikedll:
I've tested this with docker on bionic (18), focal (20), and jammy (22). By "tested", I mean I just observed that the make went down the correct path. Only jammy uses the new dependency (libdc1394-dev). I allowed the jammy build to run "make install" to completion, and the `version` cmd worked with output:

```
gocv version: 0.32.0
opencv lib version: 4.7.0
```

I also tested this with docker on debian bullseye, and observed the older deps make path being used.


deadprogram:
Thank you very much for doing this @mikedll it should help quite a few people get started! Now merging.

PR #1079, opened 2023-05-15
Deallocate Mats in feature-matching example.

mikedll:
I wrapped this example, with some adjustments, in a web server. Lo and behold my web server was leaking Mats. I tested this commit with the matprofiler and it showed 0 leaks so I think I got them.


deadprogram:
Great catch @mikedll thank you very much for the fix. Now merging.

PR #1084, opened 2023-06-13
Adds support for PCACompute

RyanHeacock:
This PR adds support for PCACompute via wrapping one of the opencv versions with this function signature. A smoke test has been added that calls the wrapped function and verifies that the parameters are being passed correctly.

In opencv there are four versions with this function signature, I have chosen to wrap (what I think is) the most common with the most parameters.

Link to supported functionality now: https://docs.opencv.org/4.x/d2/de8/group__core__array.html#ga4e2073c7311f292a0648f04c37b73781


RyanHeacock:
The smoke test that I wrote can be run with the following command: `go test -timeout 30s -run ^TestPCACompute$ gocv.io/x/gocv`

deadprogram:
Thank you very much for the feature addition @RyanHeacock now merging.

PR #1085, opened 2023-06-15
Fixes parameter ordering for new PCACompute function

RyanHeacock:
Realized after testing this implementation that I had the parameter names inverted on this. Since they are both output parameters it doesn't impact functinality so the smoke test still worked.

This change corrects the name order so that the doc is correct.


deadprogram:
Thank you for the fix @RyanHeacock now merging.

Issue #1059, opened 2023-02-24
Raspberry Pi 4B does not need "-D ENABLE_VFPV3=ON"


hare-siterwell:
When I run `make build_raspi`, there is an error reported

```sh
Required baseline optimization is not supported: VFPV3
```

swdee:
The flag `-D ENABLE_VFPV3=ON` needs to be removed from the `Makefile` build here.

https://github.com/hybridgroup/gocv/blob/579545a364744945c08186e8feca10258c73fefd/Makefile#L121

PR #1106, opened 2023-10-07
Makefile fixes for 64bit architechtures and GoCV/OpenCV version updates

swdee:
* The `Makefile` was stuck on GoCV v0.31.0 and did not match the OpenCV version 4.8.0 so this has been updated to v0.34.0.

* In OpenCV 4.8.0 there is a major regression which slowed DNN model inference by 200%  https://github.com/opencv/opencv/issues/23941 - this has been fixed in v4.8.1

* Building on 64bit Raspberry Pi fails on cmake flag `-D ENABLE_VFPV3=ON` so a conditional switch has been added for 32/64bit builds in `install_raspi`.  This fixes https://github.com/hybridgroup/gocv/issues/1059

* The `sudo_pre_install_clean` was not cleaning up libraries on 64bit machines (x86 and arm) so clearing of those directories has been implemented.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -235,10 +235,26 @@ clean:
 
 # Cleanup old library files.
 sudo_pre_install_clean:
+ifneq (,$(wildcard /usr/local/lib/libopencv*))
 	sudo rm -rf /usr/local/lib/cmake/opencv4/
 	sudo rm -rf /usr/local/lib/libopencv*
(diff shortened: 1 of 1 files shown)


deadprogram:
Hello @swdee thanks for the PR. Also see https://github.com/hybridgroup/gocv/pull/1107

Now merging, thanks again!

PR #1110, opened 2023-10-14
Add NewSURFWithParams func

Cartermel:
Added SURF alg creation with parameters.


deadprogram:
Thank you for adding the additional function @Cartermel now merging.

PR #1124, opened 2023-10-27
facemark

xyjwsj:
facemark support & fix facemark test

### DIFF
--- a/contrib/facemarkLBF_test.go
+++ b/contrib/facemarkLBF_test.go
@@ -1,0 +1,21 @@
+package contrib
+
+import (
+	"log"
+	"testing"
+)
+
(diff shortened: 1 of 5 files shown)


contrib/facemarkLBF_test.go
```diff
@@ -0,0 +1,21 @@
...
+
+func TestNewLBPHFaceMark(t *testing.T) {
+	mark := NewLBPHFaceMark()
+	log.Println(mark)
+}
+
+func TestLBPHFaceMark_LoadModel(t *testing.T) {
+	mark := NewLBPHFaceMark()
+	log.Println(mark)
+	//mark.LoadModel("")
```
deadprogram, line 16: Did you mean to actually call this in the test? Seems like that is the case from the test name.

PR #1127, opened 2023-11-23
Makefile changes for building on Debian 12/Bookworm

swdee:
The Raspberry Pi OS since October updated to Debian version 12 (Bookworm) and requires some adjustments to the Makefile to build.

The changes are;
* Removal of package `libdc1394-22-dev`
* Package `libtbb2` has been replaced with `libtbbmalloc2`


deadprogram:
Thanks for the update @swdee now merging.

PR #1132, opened 2023-12-12
Add color conversion alias

wuhaochen:
Adding some color conversion code alias same as https://github.com/opencv/opencv/blob/850be1e0874a4881b94392773d2f1702344658ac/modules/imgproc/include/opencv2/imgproc.hpp#L536

This change helps with code readability especially for demosaicing.

I left out other alias in the hpp file for now as they seem to be addressing only different naming in YUV format.
The only caveat to this change may be that the `String` function for `ColorConversionCode` in https://github.com/hybridgroup/gocv/blob/e2db8f81cfce48d9b2c2d4ce92219a7a4a021bfe/imgproc_colorcodes_string.go can return a different name for a constant, but it should be an ok compromise.

### DIFF
--- a/imgproc_colorcodes.go
+++ b/imgproc_colorcodes.go
@@ -9,22 +9,35 @@ type ColorConversionCode int
 const (
 	// ColorBGRToBGRA adds alpha channel to BGR image.
 	ColorBGRToBGRA ColorConversionCode = 0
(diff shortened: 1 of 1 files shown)


deadprogram:
Thank you for the pull request @wuhaochen now merging.

PR #1133, opened 2023-12-12
videoio: add Retrieve function

### LINKED ISSUE #928: Add support for the VideoCapture::retrieve method.
## Description
I’m trying to capture the latest frames from an RTSP stream and doing some processing which adds delay in fetching the next frame from the stream. 
I don’t need to read all the missed frames but get the latest frame and skip the remaining from the buffer.
I am planning to use two goroutines. One go routine will continuously grab(VideoCapture::grab) the frames and not process them. 
The second go routine will be used to retrieve the latest frame and process it. 

Right now I am using VideoCapture::read in the second go routine which internally calls VideoCapture::grab. And because of this, I am getting below error during read. 
```
Assertion fctx->async_lock failed at src/libavcodec/pthread_frame.c:155
SIGABRT: abort
PC=0x7fdcb53797bb m=0 sigcode=18446744073709551610
```
I assumed this is because both the go routine are calling grab at the same time. 

Ideally, in the second go routine we should just call [retrieve](https://docs.opencv.org/3.4/d8/dfe/classcv_1_1VideoCapture.html#a9ac7f4b1cdfe624663478568486e6712) and get the latest from the buffer. However, I did not find the retrieve method in gocv. 

## Your Environment
* GoCV version used: 0.29.0
* Go version: 1.17.2 darwin/amd64

AnteWall:
This pr adds `Retrieve` method to `VideoCapture`

I haven't had time to test it with a USB camera but works well with video files togheter with `Grab` method and testing different skip arguments aswell to `Grab`

Closes #928

### DIFF
--- a/videoio.cpp
+++ b/videoio.cpp
@@ -47,6 +47,10 @@ void VideoCapture_Grab(VideoCapture v, int skip) {
     }
 }
 
+int VideoCapture_Retrieve(VideoCapture v, Mat buf) {
+    return v->retrieve(*buf);
+}
+
 // VideoWriter
 VideoWriter VideoWriter_New() {
     return new cv::VideoWriter();
--- a/videoio.go
+++ b/videoio.go
@@ -379,6 +379,14 @@ func (v *VideoCapture) Grab(skip int) {
 	C.VideoCapture_Grab(v.p, C.int(skip))
 }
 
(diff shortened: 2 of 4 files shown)


deadprogram:
Thank you very much for the code @AnteWall now merging.

PR #1145, opened 2024-01-27
Add demosaicing wrapper

l0rem1psum:
I have added the wrapper for both cuda and non-cuda version of the `demosaicing` function.

The unit test for non-cuda demosaicing is completed but I have yet to write a UT for the cuda version because there seems to be a compilation error there which I think I may not be best equipped to address that. 

Would be happy to write UT for cuda demosaicing once those errors are fixed.


l0rem1psum:
Demosaicing with CUDA supports a few additional demosaic types defined in `opencv2/cudaimgproc.hpp` as [`DemosaicTypes`](https://docs.opencv.org/4.8.0/db/d8c/group__cudaimgproc__color.html#gaa9e40b798350356f45667e609d5d82f9).

This means CUDA demosaicing can either take in `gocv.ColorConversionCode` or `cuda.DemosaicType` (not yet implemented).

In order to ensure type safety and not simply change the last parameter's type to a `int`. The best way I propose is to use a generic type. For example:

In `cuda/imgproc_colorcodes.go` (not yet existent):

```go
type DemosaicTypes int

const (
	ColorBaylerBGToBGR_MHT DemosaicTypes = 256
	// ...
)

type ColorConversionCode interface {
	DemosaicTypes | gocv.ColorConversionCode
}
```

In `cuda/imgproc.go`:

```go
func Demosaicing[Code ColorConversionCode](src GpuMat, dst *GpuMat, code Code) {
	C.GpuDemosaicing(src.p, dst.p, C.int(code), nil)
}
```

There won't be any changes to the call site, and caller can now choose what type of code to put in.

However, the current minimum supported Go version is 1.13 (as in `go.mod`) which does not support generics. So I submitted a [PR](https://github.com/hybridgroup/gocv/pull/1146) to update the version to 1.21 (or can be 1.18 for better compatibility).

deadprogram:
>  there seems to be a compilation error there which I think I may not be best equipped to address that.

Hi @l0rem1psum could you please elaborate a little more on that? Thanks.

deadprogram:
Hello @l0rem1psum I think you should rebase against the latest `dev` branch as the test failure you referred to should be corrected now thanks to https://github.com/hybridgroup/gocv/pull/1167

deadprogram:
I merged your commits into `dev` branch after adding some tests. See https://github.com/hybridgroup/gocv/commits/dev/

Thanks again for working on this @l0rem1psum

PR #1149, opened 2024-02-09
Update go to 1.22.0 opencv to 4.9.0

mmichal9:


mmichal9:
@deadprogram I was hoping to start work on updating to latest go and opencv versions. 
Looks like the issue is that the base image does not exist. This is understandable since the image is not yet released. Is this done manually?

deadprogram:
> This is understandable since the image is not yet released

That is built and pushed when there is a push to the `build-opencv-images` which I took care of yesterday. See https://github.com/hybridgroup/gocv/actions/runs/8456269846

This PR should now pass CI. so now merging. Thank you for the update @mmichal9

Issue #1151, opened 2024-02-19
MinMaxIdx function does not return the correct index.


chronicom:
The problem seems to me a typo in the code.  

```
func MinMaxIdx(input Mat) (minVal, maxVal float32, minIdx, maxIdx int) {
	var cMinVal C.double
	var cMaxVal C.double
	var cMinIdx C.int
	var cMaxIdx C.int

	C.Mat_MinMaxIdx(input.p, &cMinVal, &cMaxVal, &cMinIdx, &cMaxIdx)

	return float32(cMinVal), float32(cMaxVal), int(minIdx), int(maxIdx)
}
```

In stead of returning 
```
float32(cMinVal), float32(cMaxVal), int(minIdx), int(maxIdx)
```
I guess it should be returning 
```
float32(cMinVal), float32(cMaxVal), int(cMinIdx), int(cMaxIdx)
```

Moreover, the opencv version of MinMaxIdx  returns the indexs in the form of an array of size corresponding to the dimension of the input mat. Here the index is simply a scalar. Does it mean it is a linearized version of the nd-index?

HIGHER98:
Nice spot on the typo! Besides the name of the return type, it also seems that an array should be returned with mat.dims elements. This array should always contain at least two elements as per [the OpenCV spec](https://docs.opencv.org/4.x/d2/de8/group__core__array.html#ga7622c466c628a75d9ed008b42250a73f): 

> When minIdx is not NULL, it must have at least 2 elements (as well as maxIdx), even if src is a single-row or single-column matrix. In OpenCV (following MATLAB) each array has at least 2 dimensions, i.e. single-column matrix is Mx1 matrix (and therefore minIdx/maxIdx will be (i1,0)/(i2,0)) and single-row matrix is 1xN matrix (and therefore minIdx/maxIdx will be (0,j1)/(0,j2)).

deadprogram:
Released as part of 0.36 so now closing. Thank you!

PR #1153, opened 2024-02-27
Correct MinMaxIdx function minIdx and maxIdx

HIGHER98:
Addresses issue #1151 

This PR looks into changing return type of `minIdx` and `maxIdx` in the `MinMaxIdx` function to be an `[]int` rather than `int`, i.e.

```
func MinMaxIdx(input Mat) (minVal, maxVal float32, minIdx, maxIdx []int) {
  ...
}
```

The return arrays will have input.dims elements.

---

Example:

```
	src := NewMatWithSize(10, 10, MatTypeCV32FC1)
	defer src.Close()
	src.SetFloatAt(3, 3, 17)
	src.SetFloatAt(4, 4, 16)

	minVal, maxVal, minIdx, maxIdx := MinMaxIdx(src)
```
In the above code snippet, `minVal` will be 0, `maxVal` will be 17, minIdx will be `[0,0]`, maxIdx will be `[3,3]`

### DIFF
--- a/core.cpp
+++ b/core.cpp
@@ -660,8 +660,6 @@ void Mat_MinMaxIdx(Mat m, double* minVal, double* maxVal, int* minIdx, int* maxI
 
     for(unsigned int a = 0; a < sizeof(cMinIdx)/sizeof(cMinIdx[0]); a++) {
         minIdx[a] = cMinIdx[a];
-    }
-    for(unsigned int a = 0; a < sizeof(cMaxIdx)/sizeof(cMaxIdx[0]); a++) {
         maxIdx[a] = cMaxIdx[a];
     }
 }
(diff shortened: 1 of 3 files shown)


HIGHER98:
The Appveyor tests appear to be failing for this PR, and I'm unsure why... :confused: 

The tests seem to be failing on something which I'm pretty sure should be unaffected by changes introduced here. Also, the most recent builds on Appveyor seem to be taking >30 minutes. I can see about a month ago, most builds were compiling in ~3 minutes. I suspect this is something cache-based and possibly has something to do with what's going awry. Any help would be appreciated!


Tests are failing here 
```
=== RUN   TestOnnxMemory
Exception 0xc0000005 0x0 0x0 0x64728886
PC=0x64728886
runtime: unknown pc 0x64728886
```

PR #1158, opened 2024-03-13
Add new binding for cv::Erode.

michaelarusso:
The original gocv.ErodeWithParams left out the border value parameter. A new method gocv.ErodeWithParamsAndBorderValue is added so the existing gocv.ErodeWithParams signature is not broken.


michaelarusso:
The AppVeyor build appears to have failed due to an issue unrelated to this change.

deadprogram:
Thanks for the addition @michaelarusso now merging.

Note the new GH Actions based builds for Windows in the `dev` branch.

PR #1162, opened 2024-03-28
aruco: correct test from latest OpenCV update

deadprogram:
This PR updates the `aruco` test to correct the order change from latest OpenCV update.


deadprogram:
Closing as it will be handled in #1165

PR #1180, opened 2024-06-25
dnn: blob params

deadprogram:
This PR adds a few newer functions that were missing from GoCV due to being added more recently to the `dnn` module.

It also adds a new example that uses the YOLOv8 model for object detection.


deadprogram:
Tested and works as expected both on CPU and GPU.

Now merging.

PR #1191, opened 2024-08-06
Missing HomographyMethodRHO HomographyMethod added

Krzysztofz01:
Hello!
This is my first pull-request here. There is a link to the official **OpenCV** documentation attached to the `FindHomography` function. It mentions four supported homogpraphy methods, but in **gocv** only three of them have been implemented. I ran local tests on whether the library can work with the RHO method, and all the tests were successful. In accordance with the documentation, I propose to add the missing method.

I also noticed an inconsistency in the names of values and the type (*HomographyMethod/HomograpyMethod*), but I assume that due to compatibility with previous versions of the library, it will be hard to correct this (marking old values as deprecated and adding new ones without a typo?).

Greetings

### DIFF
--- a/imgproc.go
+++ b/imgproc.go
@@ -1822,6 +1822,7 @@ const (
 	HomograpyMethodAllPoints HomographyMethod = 0
 	HomograpyMethodLMEDS     HomographyMethod = 4
 	HomograpyMethodRANSAC    HomographyMethod = 8
+	HomograpyMethodRHO       HomographyMethod = 16
 )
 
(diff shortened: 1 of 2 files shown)


deadprogram:
Thanks for the addition @Krzysztofz01 

As far as the misspelling, I think it should be fixed. Once GoCV hits a `1.0` release then we will respect the usual Go version guarantees.

If you want to make that change in this same PR, that would be fine with me! :)

Krzysztofz01:
I have corrected the typos mentioned earlier. I think the changes are ready to be merged :)

deadprogram:
Thanks for adding this @Krzysztofz01 now merging.

Issue #1184, opened 2024-07-04
Support waitKeyEx


Thanzex:
## Description

This library supports `waitKey`, but not `waitKeyEx` or `pollKey` for that matter.

`pollKey` is already a fairly big omission, but missing `waitKeyEx` is a breaking functionality. 

Right now  `waitKey` only returns  alphanumeric  characters + symbols, but no special characters.

This is a problem since directional keys, F-keys, CTRL, SHIFT or ALT or their combination are not supported and either don't show up or report as 0.

`waitKeyEx` returns the full key code, allowing the whole keyboard to be used.

## Steps to Reproduce
```go
package main

import (
	"fmt"

	"gocv.io/x/gocv"
)

func main() {
	window := gocv.NewWindow("Test")
	defer window.Close()

	for {

		/*
		* Do something
		 */

		switch k := window.WaitKey(64); k {
		case 'q':
			return
		default:
			fmt.Println("\nPressed ", k)
		}

	}
}

```

## Your Environment

* Operating System and version: Windows 11 
* OpenCV version used: 4.10.0
* How did you install OpenCV? Build in the package directory
* GoCV version used: 0.37.0
* Go version: go1.22.4 windows/amd64

deadprogram:
Yes, please!

diegohce:
Hi all!
Just sent a pull request for adding `waitKeyEx`.

cheers!

deadprogram:
Merged. Tagged to close this on next release. Thanks @diegohce

PR #1195, opened 2024-08-13
added window WaitKeyEx support

diegohce:
as per this open issue https://github.com/hybridgroup/gocv/issues/1184


deadprogram:
Thank you for the addition @diegohce can you please add some kind of smoke test for it? Thanks!

Issue #1125, opened 2023-11-03
Support for imreadmulti?


4imothy:
I am creating a program that must read a stacked tiff file, I am able to read such an image with ```imreadmulti``` in my python equivalent but would prefer to use Go. Are there plans to support this function?

* Operating System and version: MacOS Venture 13.6
* OpenCV version used: 4.8.0
* How did you install OpenCV? homebrew, ```brew install opencv```
* GoCV version used: 0.34.0
* Go version: 1.21.1 darwin/arm64
* Did you run the `env.sh` or `env.cmd` script before trying to `go run` or `go build`? No

deadprogram:
Tagging to close this issue after next release now that #1196 has been merged into `dev`.

PR #1196, opened 2024-08-14
added imreadmulti support

diegohce:
as per open issue https://github.com/hybridgroup/gocv/issues/1125


deadprogram:
Looking good to me, thank you for the addition @diegohce now merging.

PR #1199, opened 2024-08-14
make: build non-free modules when building opencv with cuda support

deadprogram:
This PR modifies the Makefile to build non-free modules when building opencv with cuda support.


deadprogram:
This is needed in order to run all of the tests alongside the CUDA tests.

Issue #1150, opened 2024-02-19
Update to Support Additional Fonts from HersheyFont


htahaozlu:
## Description

In the current version of the library, there seems to be a limitation on font support from HersheyFont, as stated in the documentation: "Only a subset of the available Hershey fonts are supported by OpenCV." However, with the release of OpenCV 4.1.1, there have been updates related to fonts, such as the addition of loadFontData. This function enables users to load additional font data, which could be particularly useful for cases like mine where Turkish words need to be rendered accurately.

I would like to request the inclusion of support for loading fonts using the loadFontData function, as it would greatly enhance the capabilities of the library, especially for users dealing with non-standard fonts or languages.

## Steps to Reproduce

1. Refer to the OpenCV 4.1.1 documentation regarding the loadFontData function: [OpenCV FreeType2 Documentation](https://docs.opencv.org/4.1.1/d9/dfa/classcv_1_1freetype_1_1FreeType2.html)
2.Attempt to use custom fonts or render text in languages with unique character sets using the current version of the library.
3.Observe the limitations and inaccuracies in font rendering.
4.Recognize the potential benefit of incorporating the loadFontData function to overcome these limitations.

## Your Environment

* Operating System and version: macOS Sonoma version 14.3.1
* OpenCV version used: 4.9.0
* How did you install OpenCV? with brew
* GoCV version used: v0.35.0
* Go version:  go 1.20
* Did you run the `env.sh` or `env.cmd` script before trying to `go run` or `go build`?

deadprogram:
A PR that added this feature would be gladly accepted! :smile_cat:

Issue #874, opened 2021-07-24
Freetype CI error


lz1998:
## Description

I create a pull request to add freetype. [PR](https://github.com/hybridgroup/gocv/pull/873)
- CircleCI: Your tests passed on CircleCI!
- AppVeyor: build failed https://ci.appveyor.com/project/deadprogram/gocv/builds/40056732

How to pass AppVeyor check?
## Steps to Reproduce

1. https://github.com/hybridgroup/gocv/pull/873

## Your Environment

* Operating System and version: AppVeyor windows/amd64
* OpenCV version used: 4.5.x
* How did you install OpenCV? Yes
* GoCV version used: [PR](https://github.com/hybridgroup/gocv/pull/873)
* Go version: 1.13
* Did you run the `env.sh` or `env.cmd` script before trying to `go run` or `go build`? No

PR #873, opened 2021-07-18
Add freetype

lz1998:
Add [FreeType2](https://docs.opencv.org/master/d9/dfa/classcv_1_1freetype_1_1FreeType2.html).

Issue #749, opened 2020-10-12
How does GoCV PutText support Chinese？


opvexe:
for _, v := range points {
		gocv.Line(&mat, image.Point{
			X: v[0].X,
			Y: v[0].Y,
		}, image.Point{
			X: v[1].X,
			Y: v[1].Y,
		}, s.BorderColor, s.Thickness)
		// 标注圆点
		gocv.PutText(&mat,"." , image.Point{
			X: v[0].X,
			Y: v[0].Y,
		}, gocv.FontHersheySimplex, 1.5, RedColor, 8)
	}
	if len(s.Typ) != 0 {
		gocv.PutText(&mat,"中文",image.Point{
			X: points[0][0].X + 20,
			Y: points[0][0].Y + 40,
		}, gocv.FontHersheySimplex, 1.5, GreenColor, 6)
	}

deadprogram:
According to this post, non-ascii character sets are not supported: https://answers.opencv.org/question/86997/how-can-i-put-text-on-image-using-opencv-puttext-function/

:crying_cat_face: sorry about that.

opvexe:
> According to this post, non-ascii character sets are not supported: https://answers.opencv.org/question/86997/how-can-i-put-text-on-image-using-opencv-puttext-function/
> 
> 😿 sorry about that.

Thank you again. I have found another way to solve the problem of Chinese in the picture.

func (fb *TextBrush) DrawFontOnRGBA(rgba *image.RGBA, pt image.Point, content string) {
	var dy int
	lineSize := int(fb.TextWidth / int(fb.FontSize))

	// 计算行数以及拆解字符串
	lines := list.New()
	tl := fb.sliptString(content, lines, lineSize)
	dy = tl * int(fb.FontSize)

	c := freetype.NewContext()
	c.SetDPI(72)
	c.SetFont(fb.FontType)
	c.SetFontSize(fb.FontSize)
	c.SetClip(image.Rect(pt.X, pt.Y, pt.X+fb.TextWidth, pt.Y+dy))
	c.SetDst(rgba)
	c.SetSrc(fb.FontColor)

	intSize := int(fb.FontSize)

	var txtNext *list.Element
	for txt := lines.Front(); txt != nil; txt = txtNext {
		txtNext = txt.Next()
		t := txt.Value.(textLine)
		pt := freetype.Pt(pt.X, pt.Y+(t.Index+1)*intSize-int(c.PointToFixed(fb.FontSize))>>8)
[...]

PR #1200, opened 2024-08-15
Add Freetype font support to PutText()

swdee:
This PR is an update to #873 which adds [Freetype font support](https://github.com/opencv/opencv_contrib/tree/master/modules/freetype) to `PutText()`.

The original work was done by @lz1998 back in 2021 but the PR has became stale due to build issues with freetype fonts on Windows.  I have modified the original PR to limit building to Linux only as the build issues on Windows should not block this important feature which supports text for non ascii characters.

This PR stops the need for workarounds as discussed in #749 to support Chinese text or at #1150.

Merging of this PR should close off #873 and #874.

### Windows

OpenCV may have now fixed the build issues on Windows here https://github.com/microsoft/vcpkg/pull/27343 so any Windows users who require this feature can confirm that and submit another PR.

To remove the Linux only build restrictions.

1. Edit `contrib/freetype.cpp` and remove the `#ifndef _WIN32` at the top of the file and `#endif // _WIN32` at the bottom.
2. Edit `contrib/freetype.go` and `contrib/freetype_test.go` files and remove the Go build tag `//go:build linux` at the top of each file.


deadprogram:
This is great work @swdee thank you very much! Now merging.

PR #1204, opened 2024-08-23
Filestorage roadmap milestone

diegohce:
Filestorage roadmap milestone


deadprogram:
Hello @diegohce 

Thank you very much for working on this set of features!

I notice that this PR seems to have picked up bunch of commits that are in the `dev` branch. Perhaps you should have started with the latest `dev` branch?

One way to address this might be to create a new branch from `dev`, then `git cherry-pick` the specific commits for filestorage into that branch, and open a new PR from that branch to `dev`.

What do you think?

diegohce:
Hi Ron.

Don't know how to do that. I understand that commit messages got dirty but
is it necessary to do that?

deadprogram:
> is it necessary to do that?

Yes, because we need to keep the commit log clean. You might be better off just creating a new PR based on `dev` and manually adding the new files in a single new commit.

diegohce:
ok

diegohce:
done. checks are running as I write this..

PR #1205, opened 2024-08-23
Test trackbar with value fix

diegohce:
test was failing on opencv layer deprecation


diegohce:
wrong gocv branch! :facepalm:

PR #1213, opened 2024-09-02
BasicFaceRecognizer + EigenFaceRecognizer + FisherFaceRecognizer

diegohce:
Face recognizers collection with "some" smoke tests

### DIFF
--- a/contrib/face_recognizer.go
+++ b/contrib/face_recognizer.go
@@ -11,7 +11,7 @@ import (
 	"gocv.io/x/gocv"
 )
 
-type IFaceRecognizer interface {
+type FaceRecognizer interface {
 	Empty() bool
 	Train(images []gocv.Mat, labels []int)
 	Update(newImages []gocv.Mat, newLabels []int)
@@ -24,7 +24,7 @@ type IFaceRecognizer interface {
 	Close() error
 }
 
-type IBasicFaceRecognizer interface {
+type BasicFaceRecognizer interface {
 	GetEigenValues() gocv.Mat
 	GetEigenVectors() gocv.Mat
 	GetLabels() gocv.Mat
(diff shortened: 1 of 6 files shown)


contrib/face_recognizer.go
```diff
@@ -11,7 +11,7 @@ import (
 	"gocv.io/x/gocv"
 )
 
-type FaceRecognizer interface {
+type IFaceRecognizer interface {
```
deadprogram: I think a better name for the interface would be just `FaceRecognizer`

contrib/face_recognizer.go
```diff
@@ -24,6 +24,26 @@ type FaceRecognizer interface {
 	Close() error
 }
 
+type IBasicFaceRecognizer interface {
```
deadprogram: I think a better name for the interface would be just `BasicFaceRecognizer`

contrib/face_recognizer.go
deadprogram: Or if that is just too hard with the `type FaceRecognizer` already, name it `FaceRecognize` and the other interface `BasicFaceRecognize`

contrib/face_recognizer.go
deadprogram: See above additional comment.

diegohce: Yes! I forgot to change them back. Was trying to implement the interfaces more go-stylish.

deadprogram:
Looks very promising @diegohce thank you!

What do you think about inserting the godocs links to the various OpenCV functions now?

diegohce:
> Looks very promising @diegohce thank you!
> 
> What do you think about inserting the godocs links to the various OpenCV functions now?

Next PR will have the godocs links. sorry for that.

deadprogram:
Thank you @diegohce now squashed/merged.

PR #1214, opened 2024-09-03
added godocs to recognizer methods

diegohce:


deadprogram:
Thank you very much @diegohce now merging.

Issue #513, opened 2019-09-24
Implement func setMouseCallback()


SilentGopherLnx:
Trying to implement **setMouseCallback()**
I have done version for one window. Can you fix it for any windows and commit?

1)**highgui_gocv.h**
```
typedef void (*mouse_callback)(int event, int x, int y, int flags, void* param);
void Window_SetMouseCallback(const char* winname, mouse_callback callback);
```
2)**highgui.cpp**
```
void Window_SetMouseCallback(const char* winname, mouse_callback callback) {
   cv::setMouseCallback(winname, callback, NULL);
}
```
3)**NEW go-file**
```
package gocv

/*
#include <stdlib.h>
#include "highgui_gocv.h"

extern void for_cgo_mouse_callback(int event, int x, int y, int flags);

static void InsideCGO_MouseCallback(char* winname) {
	void callback(int event, int x, int y, int flags, void* param) {
		for_cgo_mouse_callback(event, x, y, flags);
	};
	Window_SetMouseCallback(winname, callback);
}
*/
import "C"

type MouseCallback = func(event int, x int, y int, flags int)
var callb MouseCallback

func init() {
	mouse_callbacks = make(map[string]MouseCallback)
}

func (w *Window) SetMouseCallback(callback MouseCallback) {
	cName := C.CString(w.name)
	C.InsideCGO_MouseCallback(cName)
	callb = callback
}

//export for_cgo_mouse_callback
func for_cgo_mouse_callback(event C.int, x C.int, y C.int, flags C.int) { // cName *C.char win C.int,
	callb(int(event), int(x), int(y), int(flags))
}

//MouseEventTypes
const EVENT_MOUSEMOVE = 0
const EVENT_LBUTTONDOWN = 1
const EVENT_RBUTTONDOWN = 2
const EVENT_MBUTTONDOWN = 3
const EVENT_LBUTTONUP = 4
const EVENT_RBUTTONUP = 5
const EVENT_MBUTTONUP = 6
const EVENT_LBUTTONDBLCLK = 7
const EVENT_RBUTTONDBLCLK = 8
const EVENT_MBUTTONDBLCLK = 9
const EVENT_MOUSEWHEEL = 10
const EVENT_MOUSEHWHEEL = 11

//MouseEventFlags
const EVENT_FLAG_LBUTTON = 1
const EVENT_FLAG_RBUTTON = 2
const EVENT_FLAG_MBUTTON = 4
[...]

PR #1220, opened 2024-09-12
Window set mouse callback

diegohce:
relates to issue https://github.com/hybridgroup/gocv/issues/513


deadprogram:
Thanks for the addition @diegohce now squash/merging.

Issue #807, opened 2021-03-01
VideoCapture with GPU


alexfcmcore:
Is it possible to have videostream mat objects handled by GPUs, or jpegs directly when read the rtsp stream?

ffmpeg can do the streaming by setting  the `-hwaccel cuda` flags

PR #1224, opened 2024-09-16
Capture from device and file with HW acceleration

diegohce:
Addresses issue https://github.com/hybridgroup/gocv/issues/807


deadprogram:
:racehorse: :car: :rocket: thank you @diegohce now merging.

PR #1227, opened 2024-09-19
added missing cuda imgproc standalone functions

diegohce:
Added functions:

AlphaComp
GammaCorrection
SwapChannels
CalcHist
CalcHist_WithParams
EqualizeHist
EvenLevels
HistEven
HistRange
BilateralFilter
BlendLinear
MeanShiftFiltering
MeanShiftProc
MeanShiftSegmentation


deadprogram:
Hello @diegohce thank you very much for another great PR.

One small problem is that I think we need both the XXX and XXXWithStream versions for each CUDA function.

Please see https://github.com/hybridgroup/gocv/blob/f684ee7148c2bc244303ff8f45c9079f755c1524/cuda/arithm.go#L13-L28 for an example.

diegohce:
> Hello @diegohce thank you very much for another great PR.
> 
> One small problem is that I think we need both the XXX and XXXWithStream versions for each CUDA function.
> 
> Please see
> 
> https://github.com/hybridgroup/gocv/blob/f684ee7148c2bc244303ff8f45c9079f755c1524/cuda/arithm.go#L13-L28
> 
> for an example.

Hi @deadprogram , no problem, my bad.
I wIl fix it asap

deadprogram:
Thank you very much for the additions @diegohce now merging.

PR #1231, opened 2024-09-27
added cuda create lookuptable and cuda split func

diegohce:
added cuda create lookuptable and cuda split func from roadmap


deadprogram:
Thank you very much for another great PR @diegohce the list of cuda functions that are not implemented yet is getting shorter thanks to you! Now merging.

PR #1233, opened 2024-10-02
cgo: static builds with ffmpeg including h264 support

deadprogram:
This PR makes a number of changes to support static builds that include ffmpeg 5.0.3 static build features.

Known to work is the `x264` support, as tested by reading an MP4 file from disk using a static build.


deadprogram:
The docker images have now all been built from the `build-opencv-images` branch (5 hours of build time!) so merging this into `dev`

PR #1235, opened 2024-10-03
FaceDetectorYN example

diegohce:
added a simple FaceDetectorYN example


deadprogram:
Thank you for the example code @diegohce now merging.

PR #1236, opened 2024-10-03
fixed func name typo

diegohce:


deadprogram:
Thank you for the quick fix @diegohce now squash/merging.

PR #1242, opened 2024-10-18
feature: Alpine static image

deadprogram:
This PR is to add a Docker image for OpenCV static builds that is based on Alpine Linux 3.20.


deadprogram:
Closing in favor of #1243

PR #1247, opened 2024-10-28
GStreamer VideoWriter example

diegohce:
This PR adds an example of how to use gocv VideoWriter with gstreamer


deadprogram:
Thank is great thank you very much @diegohce now merging.

PR #1256, opened 2024-11-30
New mat from point vector

diegohce:
This PR adds a new constructor for the Mat type

### DIFF
--- a/core_test.go
+++ b/core_test.go
@@ -3412,3 +3412,23 @@ func TestNewRotatedRect2f(t *testing.T) {
 	}
 
 }
+
+func TestNewMatFromPointVector(t *testing.T) {
+
+	img := IMRead("images/face.jpg", IMReadColor)
+	defer img.Close()
+
+	size := img.Size()
+
+	points := []image.Point{
+		image.Pt(0, 0), image.Pt(0, size[0]-1),
+		image.Pt(size[1]-1, size[0]-1),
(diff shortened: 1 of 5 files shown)


core_test.go
```diff
@@ -3412,3 +3412,23 @@ func TestNewRotatedRect2f(t *testing.T) {
...
+		image.Pt(0, 0), image.Pt(0, size[0]-1),
+		image.Pt(size[1]-1, size[0]-1),
+		image.Pt(size[1]-1, 0),
+	}
+
+	pv := NewPointVectorFromPoints(points)
+	defer pv.Close()
+
+	m := NewMatFromPointVector(pv, false)
+	defer m.Close()
```
deadprogram, line 3433: Maybe add some test that `m` is not empty?

core_test.go
```diff
@@ -3412,3 +3412,23 @@ func TestNewRotatedRect2f(t *testing.T) {
 	}
 
 }
+
+func TestNewMatFromPointVector(t *testing.T) {
+
+	img := IMRead("images/face.jpg", IMReadColor)
```
deadprogram: Not sure why you need `img` in this test. You could just set `size` to some specific value.

core_test.go
diegohce: fixing the test as I write this.

diegohce:
Closed in favor of a cleaner PR.

PR #1259, opened 2024-12-02
added new Mat constructor and tests

diegohce:
This PR adds NewMatFromPointVector() function. Also a missing test for NewMatFromPoint2fVector()


deadprogram:
Thanks @diegohce !

PR #1271, opened 2025-01-16
Fixed wrong links in cmd Readme

shivivats:
Fixed wrong links for face detection from url and find chessboard sample programs.


diegohce:
Thanks for the corrections @shivivats

Issue #1156, opened 2024-03-08
Compilation yields "/bin/sh: line 0: cd: OLDPWD not set" on MacOS


serge-hulne:
## Description
```
 make install
rm -rf /tmp/opencv
mkdir /tmp/opencv
cd /tmp/opencv
curl -Lo opencv.zip https://github.com/opencv/opencv/archive/refs/tags/4.8.1.zip
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 92.1M    0 92.1M    0     0   726k      0 --:--:--  0:02:09 --:--:--  836k
unzip -q opencv.zip
replace opencv-4.8.1/.editorconfig? [y]es, [n]o, [A]ll, [N]one, [r]ename: A
curl -Lo opencv_contrib.zip https://github.com/opencv/opencv_contrib/archive/refs/tags/4.8.1.zip
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 58.6M    0 58.6M    0     0   582k      0 --:--:--  0:01:43 --:--:--  748k
unzip -q opencv_contrib.zip
replace opencv_contrib-4.8.1/.editorconfig? [y]es, [n]o, [A]ll, [N]one, [r]ename: A
rm opencv.zip opencv_contrib.zip
cd -
/bin/sh: line 0: cd: OLDPWD not set
```

## Environment

MacOS : 14.3.1 (23D60) 
Apple silicon M1
brew install opencv
/opt/homebrew/Cellar/opencv/4.9.0_4
go version go1.21.3 darwin/arm64

```
sh env.sh   
This script is no longer necessary and has been deprecated.
See the Custom Environment section of the README if you need to customize your environment.
```

alan890104:
me too, i encounter the same problem with my m2 air

deadprogram:
You should install OpenCV using Homebrew on macOS.

See https://gocv.io/getting-started/macos/

PR #1276, opened 2025-02-06
feat: Add OS check to Makefile install

SoulPancake:
This can help avoid issues like #1156 

This can be extended to add other cases like such which might commonly cause confusion to folks who might have not visited the guide and went code-first. Even though that's not the best practise.
What are your thoughts @deadprogram

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -276,7 +276,14 @@ endif
 endif
 
 # Do everything.
-install: deps download sudo_pre_install_clean build sudo_install clean verify
+install:
+ifeq ($(shell uname -s),Darwin)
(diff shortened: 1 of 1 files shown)


deadprogram:
I think it is a good idea to avoid possible confusion for macOS in this case. More generally, not sure. But this PR seems like a good thing to me.

I changed the branch for it to `dev` as mentioned here: https://github.com/hybridgroup/gocv/blob/release/CONTRIBUTING.md#how-to-use-our-github-repository

Now squash/merging. Thanks @SoulPancake

PR #1280, opened 2025-02-19
feat: Add BRISK with params support

SoulPancake:
One of the TODOs in the repo for adding BRISK with params
Unsure if the testing is sufficient, open to discussion and suggestions for this
@deadprogram


SoulPancake:
Also I made the assumption that we should open this to the dev branch first, like for my previous PR 
@deadprogram

deadprogram:
Looks great @SoulPancake thank you very much for the addition! Now squash/merging.

PR #1291, opened 2025-03-16
feature: add exception handling function and try/catch blocks

deadprogram:
This PR is to add exception handling functions and try/catch blocks around calls to OpenCV functions to prevent unexpected exceptions from crashing the process.

The way that it works is most functions that were previously not returning anything, now return an error. The error will be populated with any exception caught by the calls to OpenCV that would have previously resulted in a crash due to that exception. These errors are very reliable and thread-safe.

```go
	img := gocv.NewMat()
	defer img.Close()

	dest := gocv.NewMat()
	defer dest.Close()

        // previously this would crash due to exception
	err := gocv.CvtColor(img, &dest, ColorBGRAToGray)
        // info in error comes directly from failed function call
	if err == nil {
		t.Error("Expected exception in test")
	}
	if len(err.Error()) == 0 {
		t.Error("Expected exception message in test")
	}
```

In addition, this PR introduces some new functions to try to figure out if an exception has occured after calling functions that already returned something else besides an error:

```go
func LastExceptionError() error
```

Here is how you would use it:

```go
net := ReadNet("googlenet-9.onnx")
if err := gocv.LastExceptionError(); err != nil {
       t.Errorf("There was an exception: %v", err)
}
...
```

However these new functions are not thread safe. They only exist for the case that the user needs to check for exceptions after calling one of the function that does not currently return an error.

Perhaps in the future the functions will make a breaking change to also return errors, but that is not part of this PR.

In short: the errors returned from a function call with exception info should be very reliable. The other new functions, less so.

### DIFF
--- a/aruco.cpp
+++ b/aruco.cpp
@@ -1,11 +1,21 @@
 #include "aruco.h"
 
 ArucoDetector ArucoDetector_New() {
-    return new cv::aruco::ArucoDetector();
+    try {
+        return new cv::aruco::ArucoDetector();
+    } catch(const cv::Exception& e){
+        setExceptionInfo(e.code, e.what());
+        return NULL;
+    }
 }
 
 ArucoDetector ArucoDetector_NewWithParams(ArucoDictionary dictionary, ArucoDetectorParameters params) {
-    return new cv::aruco::ArucoDetector(*dictionary, *params);
+    try {
+        return new cv::aruco::ArucoDetector(*dictionary, *params);
+    } catch(const cv::Exception& e){
+        setExceptionInfo(e.code, e.what());
+        return NULL;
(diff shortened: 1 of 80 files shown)


deadprogram:
I've added try/catch around the `contrib` and `cuda` subpackage functions now.

Ready for review!

deadprogram:
There is a test in the main package, another one in the `contrib` subpackage and another in the `cuda` subpackage that test the exception handling on Linux, macOS, and also Windows.

The fact that the tests are passing seems to indicate that all is working as expected! :tada:

SoulPancake:
That's a gigantic PR to review :p 
Hopefully the tests cover every possible edge case

PR #1304, opened 2025-04-10
Add cv::stereoRectify binding.

### LINKED ISSUE #1303: Add support for cv::stereoRectify

michaelarusso:
Note: The original enumerations for calibration flags are for the FishEye model but do not reference FishEye or "FE" in the name like they are in opencv source code. Because of this, I included "Pinhole" in the new enumeration naming for clarity that the flags are used with calibration functions for the pinhole model. 

Closes #1303


deadprogram:
Thank you very much for another useful addition @michaelarusso now merging.

PR #1317, opened 2025-06-03
imgproc: update DilateWithParams function to use int for borderType parameter

dwSun:
imgproc: update DilateWithParams function to use int for borderType parameter


deadprogram:
@dwSun not sure what is the problem you are trying to solve here? 

You can cast an `int` to `gocv.BorderType` if you need to.

dwSun:
The function declarations for DilateWithParams and ErodeWithParams are inconsistent, which is quite confusing

deadprogram:
I think I did not understand the problem you were describing @dwSun 

This has probably now been addressed by https://github.com/hybridgroup/gocv/pull/1331

PR #1348, opened 2025-12-18
Added in missing ColorMaps

swdee:
This PR adds in the [missing Colormaps](https://github.com/opencv/opencv/blob/0ca4117c272611e0576b272b0f93fa01e0d8361a/modules/imgproc/include/opencv2/imgproc.hpp#L4597-L4621).

Just for fun, here is how they all look!

gocv-all-colormaps


deadprogram:
Thanks @swdee :art: now merging.

PR #1367, opened 2026-04-23
Add getter for ushort

michaelarusso:


deadprogram:
Thanks for adding this @michaelarusso now merging.

PR #1368, opened 2026-04-30
Fix binding for cv::getRectSubPix

michaelarusso:
Updated the GetRectSubPix binding to use a float center (Point2f) end-to-end instead of integer coordinates, matching OpenCV’s API and intended behavior. The previous integer center lost precision at the boundary between Go and C++.

**Why int casting defeats the purpose:**
GetRectSubPix is specifically for sub-pixel sampling. Casting center coordinates to int truncates fractional values, so a center like (200.5, 172.5) becomes (200, 172). That removes sub-pixel precision and effectively turns the call into whole-pixel extraction, which defeats why this function exists.

**Note:** 
This is a breaking API change: GetRectSubPix now accepts Point2f for center instead of image.Point. Existing clients must update call sites to pass Point2f values.

### DIFF
--- a/imgproc.cpp
+++ b/imgproc.cpp
@@ -902,10 +902,10 @@ OpenCVResult Resize(Mat src, Mat dst, Size dsize, double fx, double fy, int inte
     }
 }
 
-OpenCVResult GetRectSubPix(Mat src, Size patchSize, Point center, Mat dst) {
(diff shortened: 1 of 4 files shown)


deadprogram:
Oops sorry @michaelarusso I missed the notification on this.

Yes, it is breaking, but the original code was incorrect I see now. As such, merging now. Thanks you for the fix!

